PR #1, opened 2015-09-25
a compulsory type conversion is not needed at all!

Snake1999:


MagicDroidX:
the method is already "T getOwner()" in this branch

PR #6, opened 2015-10-11
addded some packets handling and events

Creeperface01:


Creeperface01:
there are some mistakes in "import" so then i fix them

Creeperface01:
maybe Snowball NBT tag is bad

MagicDroidX:
That's great!!!! :D

Creeperface01:
i write other packets later (some of them today)

MagicDroidX:
That's awesome!

Creeperface01:
thx :D and can you check snowbal NBT tag? in play.php

MagicDroidX:
There are alot to check :D

Creeperface01:
Yes but maybe better is make server able to start and try it :D

MagicDroidX:
I suggest you to use Intellj IDEA for coding. It is better to check out the codes

Creeperface01:
I make this in github editor xD because i have not java IDE so which IDE should i use? Eclipse?

Snake1999:
@Creeperface01 I suggest you to use Intellj IDEA . http://www.jetbrains.com/idea/

Creeperface01:
Ok im using phpstorm for PHP so i hope it is the best IDE too

MagicDroidX:
Yes. And i m very sorry but it can not simply do replacing work.

MagicDroidX:
So it's pretty BAD to remove your codes but it seems better to write it later I think :/.

Creeperface01:
Why? Im new in java so my code is not without errors i think but there are not too many errors right?

MagicDroidX:
Yes! Half of the codes of `Player.java` are **errors**. That's because we haven't done every class of pocketmine, so you know. And also java is not simply like php. It's pretty different while the syntax is similar.

Creeperface01:
Yes about classes it is right but you have almost same classes like pocketmine

Creeperface01:
So can i add blocks or items?

MagicDroidX:
That's the best, and don't forget to use IDEA :D.

Creeperface01:
Ok :D

Creeperface01:
So can i try to upload right file? :D

MagicDroidX:
@RamdomAltThingy LMAO. Copy the codes is probably not right. But why this project is a redistribution of PocketMine. If wirte the same method name and class name is redistribution, then pocketmine is also a redistribution of bukkit and even craftbukkit. PocketMine **copy** java codes in bukkit and even craftbukkit.

PR #8, opened 2015-10-12
Impl Player.kick()

caoli5288:


MagicDroidX:
The packet must have a no-argument constructor for reflection.

caoli5288:
Ok. I'll notice it later.

PR #10, opened 2015-10-13
Added more effects (hunger is not working)

PeratX:
I will implement effects for effects later.


PeratX:
格式不太好，别介意=。=

MagicDroidX:
It's ok, i will format these.

PR #17, opened 2015-10-14
Simply scheduler implements

caoli5288:
This is a more simply and gracefully implements. Use Java's native Thread Pool Executor to handle async task.


iNevet:
Thank you :)

PR #29, opened 2015-10-24
Added enchantments

Creeperface01:
but actually i don't know how you NBT works so i add methods to item class later or you can


Creeperface01:
why?

Creeperface01:
OK you added it

PR #58, opened 2015-11-23
have fun with breaking blocks!

Angelic47:
Now bug was fixed! Have fun with placing and breaking blocks!


MagicDroidX:
@caoli5288 Nice Packet :D

PR #59, opened 2015-11-23
Flowers and FenceGates

Snake1999:
Still something todo...

### DIFF
--- a/src/main/java/cn/nukkit/block/FenceGate.java
+++ b/src/main/java/cn/nukkit/block/FenceGate.java
@@ -1,0 +1,98 @@
+package cn.nukkit.block;
+
+import cn.nukkit.Player;
+import cn.nukkit.item.Item;
+import cn.nukkit.item.Tool;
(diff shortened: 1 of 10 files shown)


src/main/java/cn/nukkit/block/FenceGate.java
```diff
@@ -0,0 +1,98 @@
...
+    public int[][] getDrops(Item item) {
+        return new int[][]{
+                {this.id, 0, 1}
+        };
+    }
+
+    @Override
+    public boolean onActivate(Item item, Player player) {
+        int[] faces = new int[]{3, 0, 1, 2};
+        this.meta = (faces[(player != null) ? player.getDirection() : 0] & 0x03) | ((~this.meta) & 0x04);
```
Snake1999: Here is a bug, when a player closes the fence gate, the fence gate will change its direction. and mcpe client is so laggy that slows debugging. Fix it later.

PR #87, opened 2015-11-29
fix banlist: java.lang.ArrayIndexOutOfBoundsException: 0

Pub4Game:
fix banlist: java.lang.ArrayIndexOutOfBoundsException: 0


Pub4Game:
It's hard to understand Java if you were doing always only php...

PR #93, opened 2015-11-30
Fix some IDE can't compiling

qq505916481:
Using 'UTF-8' encoding to copy filtered resources.


MagicDroidX:
Would u plz remove useless tabs?

qq505916481:
dd

PR #121, opened 2015-12-05
Update FallingSand.java

rcsupermanjob:
Drop！


MagicDroidX:
I wonder if there need more specific judge with blocks including Bricks

PR #136, opened 2015-12-08
implementation /kill

Pub4Game:
This is my first treasure in Nukkit :) If something wrong, please tell me about it


caoli5288:
Nice :+1:

PR #141, opened 2015-12-10
"colorize" method for TextFormat

fromgate:
Now plugin develoers will be able to use TextFormat.colorize ("&cred text &gold text");


Tee7even:
Why not call it like in Bukkit - translateAlternateColors? Let's not produce too much differeces.

fromgate:
TextFormat is not a ChatColor too :) 
"colorize" is shorter and totally describes the method's action. And I think there's no reason to fully clone translateAlternateColors - everyone uses the "&" symbol when defining colors.

Tee7even:
So, The difference between ChatColor and TextFormat is the reason to make more differences? What's that "easy to rewrite plugins" is all about? Of course, Nukkit API mimics PocketMine and not Bukkit, but for methods that originated from Bukkit, it would make sense to call them as they called in Bukkit.

PR #145, opened 2015-12-10
implementation Melon Block

Pub4Game:


src/main/java/cn/nukkit/block/Melon.java
```diff
@@ -0,0 +1,45 @@
...
+*/
+
+public class Melon extends Transparent {
+
+    public Melon() {
+        this(0);
+    }
+
+    public Melon(int meta) {
+        super(meta);
```
MagicDroidX: If you found codes like this: `public function __construct(){}` in PocketMine's block classes, use `super(0);` instead of `super(meta);`

PR #146, opened 2015-12-10
Redstone Block

Pub4Game:


Snake1999:
Oh, your forgot getResistance() . Just add these:
```
    @Override
    public double getResistance() {
        return 30;
    }
```
The number 30 is copied from http://minecraft.gamepedia.com/Block_of_Redstone.

Pub4Game:
Ohh, thanks, I'll fix it

src/main/java/cn/nukkit/block/Redstone.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return "Redstone Block";
+    }
+
+    @Override
+    public int[][] getDrops(Item item) {
+        if (item.isPickaxe()) {
```
MagicDroidX: Use this: `if (item.isPickaxe() && item.getTier() >= Tool.TIER_WOODEN) {`

PR #163, opened 2015-12-14
Rename ButtonClickSound, add real NoteBoxSound, fix player food level

boybook:
Rename ButtonClickSound, add real NoteBoxSound, fix player food level when respawn


src/main/java/cn/nukkit/level/sound/ButtonClickSound.java
```diff
@@ -0,0 +1,18 @@
+package cn.nukkit.level.sound;
+
+import cn.nukkit.math.Vector3;
+import cn.nukkit.network.protocol.LevelEventPacket;
```
boybook: 0

PR #186, opened 2015-12-19
Add ability to save resource to defined file

fromgate:
If you trying to save resource file using method plugin.saveResource ("lang/russian.lng", false); you can save it only to "<PluginName>/lang/russian.lng". 
But I think we need ability to save resource to file located anywhere (without keeping file path and name).

For example:
File out = new File ("current-language.lng");
plugin.saveResource ("lang/russian.yml", false,out);

### DIFF
--- a/src/main/java/cn/nukkit/plugin/PluginBase.java
+++ b/src/main/java/cn/nukkit/plugin/PluginBase.java
@@ -184,6 +184,11 @@ abstract public class PluginBase implements Plugin {
 
     @Override
     public boolean saveResource(String filename, boolean replace) {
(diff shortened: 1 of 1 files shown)


caoli5288:
BUGGY code. It does not check if the out file's folder exists. The out file may outside of data folder! I suggest REVERT this commit. Or change this API to saveResource(String, String) to ensure out file inside data folder.

fromgate:
I'm not sure why it must be forbidden to save resource outside the plugins' folder? 
If it really required I can add checks, can you explain why it needed?


Anyway I think saveResource method must throw exceptions instead of returning true or false - developer must know what's going wrong :)

PR #187, opened 2015-12-19
redstone (part 2), would u like it ?

Angelic47:
now part 2 avaliable. when you break redstone wires, redstone torch, all redstones will update.
but still need to do alot.


Pub4Game:
Good Job :+1:

Snake1999:
Well done :+1:

fromgate:
You're great! :+1:

Angelic47:
updated! now redstone can update when you break it.

PR #202, opened 2015-12-23
Change Plugin.saveResource()

caoli5288:
More argument check. And prevent create file out of data folder.


caoli5288:
@fromgate That's right.

PR #210, opened 2015-12-25
Exp and Exp bottles (ABLE TO MERGE)

Snake1999:
For potions, It seems like potion system requires food system done. So I'll do this PR only in exp and exp bottles.


Creeperface01:
Good job but i think my method for adding XP is better

PR #294, opened 2016-01-08
Fixed block place Quartz

Pub4Game:
image


Angelic47:
Using 0bxxxxxxxx is also ok :3

PR #304, opened 2016-01-09
temporal fix for #266

Creeperface01:


MagicDroidX:
Nooooooo

Creeperface01:
Lang.ini is wrong right?

MagicDroidX:
Of course not

Creeperface01:
So what's wrong?

Creeperface01:
Transtalion method?

Angelic47:
I fixed this, now I'm uploading.
It caused by e.getMessage(), which returns null and generated nullpointexception. I changed it to e.toString();
But as you see, there e.getMessage() is in a catch, so there must be somethings wrong, not only here.

Creeperface01:
Yes but this cause level lag

Creeperface01:
And doublechestInventory constructor is wrong because left and right inventory is defined late

Angelic47:
I know, but it's no use to fix e.getMessage() because there must have another bug.
The reason why I fixed this NPE is to get the right error message, so we can see what went wrong with level.

Creeperface01:
DoubleChestInventory constructor. This is the bug.

Creeperface01:
Right and left inventory must be defined before parent constructor. So idk how to fix it.

Angelic47:
Yes, you are right. DoubleChestInventory has a serious bug, which can cause many problems.

Creeperface01:
Do you know how to fix it?

Angelic47:
Sorry, I don't know how to fix it either... But I will try later, now I'm writting redstone support

Creeperface01:
Ok

PR #319, opened 2016-01-10
Update Sapling.java

ItsLucas:


src/main/java/cn/nukkit/block/Sapling.java
```diff
@@ -59,7 +61,38 @@ public boolean canBeActivated() {
...
+    }
+
+    public int onUpdate(int type) {
+        if (type == Level.BLOCK_UPDATE_NORMAL) {
+            if (this.getSide(0).isTransparent()) {
+                this.getLevel().useBreakOn(this);
+                return Level.BLOCK_UPDATE_NORMAL;
+            }
+        } else if (type == Level.BLOCK_UPDATE_RANDOM) { //Growth
+            if (new cn.nukkit.utils.Random().nextRange(1, 7) == 1) {
```
Angelic47: 你也可以先import cn.nukkit.utils.Random，然后new出一个random对象来，之后复用，这样代码看起来比较整洁而且运行效率更高

U also can using "import cn.nukkit.utils.Random", then new a Random object and reuse it. That would be looks nice and more faster.

PR #337, opened 2016-01-13
Chat: minimize GC throughput

blogcin:
I think it is better.

### DIFF
--- a/src/main/java/cn/nukkit/event/player/PlayerChatEvent.java
+++ b/src/main/java/cn/nukkit/event/player/PlayerChatEvent.java
@@ -21,25 +21,15 @@ public class PlayerChatEvent extends PlayerMessageEvent implements Cancellable {
 
     protected Set<CommandSender> recipients = new HashSet<>();
 
-    public PlayerChatEvent(Player player, String message) {
-        this(player, message, "chat.type.text", null);
-    }
-
-    public PlayerChatEvent(Player player, String message, String format, Set<CommandSender> recipients) {
+    public void sendMessage(Player player, String message) {
         this.player = player;
         this.message = message;
+	this.format = "chat.type.text";
 
-        this.format = format;
-
-        if (recipients == null) {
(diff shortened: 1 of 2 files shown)


src/main/java/cn/nukkit/event/player/PlayerChatEvent.java
```diff
@@ -21,25 +21,15 @@ public static HandlerList getHandlers() {
...
 
-    public PlayerChatEvent(Player player, String message) {
-        this(player, message, "chat.type.text", null);
-    }
-
-    public PlayerChatEvent(Player player, String message, String format, Set<CommandSender> recipients) {
+    public void sendMessage(Player player, String message) {
         this.player = player;
         this.message = message;
+	this.format = "chat.type.text";
```
Angelic47: Ohh sorry but, if you do this, will setFormat() doesn't work?

blogcin: What that mean? 
I think setFormat() will work! :)

Angelic47: In src/main/java/cn/nukkit/event/player/PlayerChatEvent.java, line 27.
When you call to sendMessage, it would this.format = "chat.type.text";
So in my opinion, format will always be "chat.type.text" and no others, I'm not sure if I'm right, but maybe here need to check if format is changed and you can't assign a const string to it directly.

blogcin: I just changed constructor into method. I didn't change essence of codes.
It is similar with previous codes, and chat is work successfully!

grzegorz2047: this.format will be always overridden so even if you use setFormat() then sendMessage will be used after that and overridde changed format. So you should let as it is or check if this.format is empty then this.format = "chat.type.text"; otherwise let dev to use his format.

xpyctum: Bad practice.

PR #367, opened 2016-01-16
Allow to check who killed player.

grzegorz2047:


Snake1999:
Oh, you needn't do this. To get the killer, you can simply get the EntityDamageEvent which cause this player's death, and get data from it :3

grzegorz2047:
I think,  it might not be that obvious  xd It's handy to have such field. It's bukkit port friendly xd If you think that it's redundant then you can close it.

Snake1999:
For me, I think most of our plugin developers comes from pocketmine, and what Nukkit does the best is, to let them not be confused about API. Adding another way of API might confuse everyone. :/

I thinks it's redundant, but I should wait for our team to discuss about it :D

grzegorz2047:
Ok then. Btw.  Pocketmine's api is inspired by bukkit api as I remember xd And I was trying to get the killer like in bukkit,  but there was not such thing like getKiller so I added here this method.

grzegorz2047:
or at least we can add it to playerdeathevent direcly  like e. getKiller()  if it would has more sense.  less painful to port bukkit plugin xd

MrPowerGamerBR:
@grzegorz2047 exactly.

I'm porting some Bukkit plugins to Nukkit, and I got stuck on that part, because I don't know how to get the Killer from a PlayerDeathEvent.

While PocketMine: MP developers doens't use getKiller, Bukkit developers uses it on the PlayerDeathEvent, and, in my opinion, is more convinent to have it.

Angelic47:
In my opinion, it would easy to port Bukkit plugins to Nukkit, so I think it's a good idea to get the Killer from a PlayerDeathEvent.
It won't slow down the server, and plugin developers need a faster and easier way to get the killer. So this maybe a helpful function in plugin development.

ItsLucas:
Pretty darn good
let's have it

fromgate:
I'm totally agree with @Angelic47 
PlayerDeathEvent is used often than EntityDeathEvent in common plugins.
So if it possible to save sometime for plugin developers it will be good :)

PR #370, opened 2016-01-17
Position class removed

fromgate:
Position class was united with Location. Now there's no two similar but separated classes. Everywhere where Position was used, we will use Location. 

Why we need this change:
Now we have a three way that points to coordinates:
* Location
* Position
* Vector

As plugins developer I can say that it is really terrible to guess is object is vector, location or position. To guess where I need to use locations and where positions. So best way is to unite this two classes to one. 

So we need to use: 
* Location everywhere,
* Vector3 only in classes where level is known (Level and Chunk classes).


MagicDroidX:
:+1:

Snake1999:
Great idea :+1:

grzegorz2047:
:+1:

xpyctum:
-1

PR #391, opened 2016-01-23
implementation EnchantCommand :D

Pub4Game:


src/main/java/cn/nukkit/command/defaults/EnchantCommand.java
```diff
@@ -0,0 +1,53 @@
...
+        if (args.length < 2) {
+            sender.sendMessage(new TranslationContainer("commands.generic.usage", this.usageMessage));
+            return true;
+        }
+        Player player = sender.getServer().getPlayer(args[0]);
+        if (player == null) {
+            sender.sendMessage(new TranslationContainer(TextFormat.RED + "%commands.generic.player.notFound"));
+            return true;
+        }
+        int enchantId = Integer.parseInt(args[1]);
```
Angelic47: Add a try... catch..., or will get an exception when args are not in number format. Just leave a note for myself.

Pub4Game: Thank you, I'll know

PR #394, opened 2016-01-23
Minor additions to Location and Position classes: it will be useful

fromgate:
Fix Location and position methods will return Location and Positions :)
Added getBlock() method  to Position

### DIFF
--- a/src/main/java/cn/nukkit/level/Location.java
+++ b/src/main/java/cn/nukkit/level/Location.java
@@ -1,6 +1,7 @@
 package cn.nukkit.level;
 
 import cn.nukkit.math.Vector3;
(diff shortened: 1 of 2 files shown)


xpyctum:
:+1:

MrPowerGamerBR:
```
[ERROR] /home/travis/build/Nukkit/Nukkit/src/main/java/cn/nukkit/entity/FallingSand.java:[159,16] getBlock() in cn.nukkit.entity.FallingSand cannot override getBlock() in cn.nukkit.level.Position

  return type int is not compatible with cn.nukkit.block.Block
```

Yeah, something needs to be fixed before merging it.

fromgate:
OOps!

PR #395, opened 2016-01-23
Minor additions to Location and Position classes: it will be useful

fromgate:
Fix Location and position methods will return Location and Positions :) …
Added getLevelBlock() method  to Position

### DIFF
--- a/src/main/java/cn/nukkit/level/Location.java
+++ b/src/main/java/cn/nukkit/level/Location.java
@@ -1,6 +1,7 @@
 package cn.nukkit.level;
 
 import cn.nukkit.math.Vector3;
+import cn.nukkit.utils.LevelException;
 
 /**
  * author: MagicDroidX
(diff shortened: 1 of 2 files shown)


Snake1999:
:+1:

MagicDroidX:
Good but, why? Position is the parent of Location, we don't have to override the methods with the same codes.

fromgate:
@MagicDroidX 
Location.add (x,y,z) returns... Vector3.

Without this fixes you need every time use something like this:
```
Vector3 v = loc.add(1,2,3);
newLoc = new Location (v.x,v.y,v.z,loc.yaw,loc.pitch,loc.level);
```

Other way is to modify Vector3 class  - so it will not create new vector object for add/substract methods (modify coordinates and return self) - Like it works in bukkit.

But as I can understand you translated vectors from PocketMine and this must be left untouched.

fromgate:
Same for position. Positions are parent for Location but now pos.add(x,y,z) returns Vector3. I think it must return Position.

fromgate:
> Good but, why? Position is the parent of Location, we don't have to override the methods with the same codes.

Code is not same. Did you forget about Yaw pitch in location? And level in position? :)

MagicDroidX:
:D

PR #438, opened 2016-01-31
Working NoteBlock.

xpyctum:


MrPowerGamerBR:
:+1:

Pub4Game:
@MrPowerGamerBR it doesn't work ;/

MrPowerGamerBR:
@Pub4Game oh... then :+1: for trying... I guess...

ProjectInfinity:
Perhaps reject this then? We have no need for broken implementations, do we?

xpyctum:
It will fix now, wait commit :)

ProjectInfinity:
You should be amending your latest commits if you made a mistake or didn't include a change that you meant to.

https://www.atlassian.com/git/tutorials/rewriting-history/git-commit--amend/

We need some contributor guidelines for Nukkit. Tracking down commits is really difficult thanks to the insane amount of poorly described and often spammy commits. We often get commits saying "Updated block.java" etc, which is not very descriptive and definitely doesn't help you figure out the offending commit when looking through history whenever something breaks.

ProjectInfinity:
The above commits just explain my issue with this in a nutshell. If this was accepted we suddenly have ALL of these commits put into the master branch with NO description whatsoever.

In order:
- Fixed all. Works fine.
- Fix.
- Oops
- Working Noteblock

This is the example of a messy pull request and even worse commits. The fact that you want to implement something is *great*, just please do it in a clean way. All of your changes could be combined into a single commit just called "Implements NoteBlock" and it would be descriptive and add no garbage to the repo history.

xpyctum:
@ProjectInfinity Sorry.

PR #448, opened 2016-02-01
Fix spruce trees

fromgate:
Fix the top-block of spruce trees. Now top block will be always a leaves block

Bug screenshot:

spruce-trees-top-block


Pub4Game:
:+1:

MrPowerGamerBR:
:+1: I always wondered if that was a bug or MCPE had trees like that (instead of trees like the PC version)

PR #450, opened 2016-02-02
Better ways to do configuration

berry64:
NO MORE YAML! instead uses this awesome thing called JSON...

uses JSON-lib
well.... yeah, its not like i can write my own JSON reader......


fromgate:
It looks like Config class supports JSON: https://github.com/Nukkit/Nukkit/blob/master/src/main/java/cn/nukkit/utils/Config.java#L26

I did not test it yet, but it includes the code :)

berry64:
yes it does contain JSON but 1) it's not that efficient (idk actually, but it does look like a lot more coding to me)
2) it limits itself to only config.yml/.json

Snake1999:
It already supports json. Just see how banned-ips.json and banned-players.json works. But thanks for your works

PR #465, opened 2016-02-05
Fix getDirectionVector() return wrong direction vector 

fromgate:
Hello! I found that getDirectionVector returns wrong vector.

How to reproduce:
```
    @Override
    public boolean onCommand(CommandSender sender, Command command, String label, String[] args) {
        if (!command.getName().equalsIgnoreCase("test")) return false;
        Player player = sender instanceof Player ? (Player) sender : null;

        Block[] blocks = player.getLineOfSight(10);
        for (Block b : blocks)
            b.getLevel().setBlock(b, Block.get(Block.GLASS));
    }
```

Here is demonstration of bug: 
https://youtu.be/epKJ4kfWslA

How bug was fixed:
1. getDirectionMethod() ignores fact that yaw/pitch in minecraft is "skewed". You can read more about it here: https://bukkit.org/threads/tutorial-how-to-calculate-vectors.138849/
2. I moved getDirectionLocation to Location class, because it could be useful in locations too. 

Here is result (test command code was not changed): https://youtu.be/2wQ_8xo1gTM

P.S. 
I think getDirectionPlane is buggy too. But I'm not sure what is goal of it, so I cannot fix it. Please check it.


Pub4Game:
:+1:

Snake1999:
:+1:

PR #470, opened 2016-02-06
Fix tiles not loading on premade maps.

xxFlare:


Snake1999:
That's nice. And remember mojabg has renamed "tile" to "block entity" :D

xxFlare:
@Snake1999 Oh.. But they haven't changed anything in them right?

Snake1999:
Yes you're right :D

xxFlare:
Thanks for merging! Now my players can see chests.

Now time to figure out why teleports on PlayJoinEvent won't work..

PR #482, opened 2016-02-10
Support For v0.13.1 & v0.13.2

chaunhan:


xxFlare:
This won't even work & there is no reason of why we would want to support two versions.

They will send different packets and Nukkit will get confused.

Creeperface01:
@chaunhan this is not php

Snake1999:
Next time, you can send 1 pull request instead of many :3

PR #486, opened 2016-02-11
Added SimpleConfig class

fromgate:
SimpleConfig is a class that created to made working with config files easier.
It based on built-in Nukkit Config class (using this class to save/load operations).

```
public class TestCfg extends SimpleConfig {
    public TestCfg(PluginBase plugin) {
        super(plugin);
    }

    @Path (value = "test.string")
    public String testString = "test string";

    @Path (value = "test.integer")
    public int testInt = 10;

    @Path (value = "test.list")
    public List<String> testStringList = new ArrayList<String>(Arrays.asList("test 1", "test 2", "test 3"));
}
```

To use config now must create a TestCfg object:

```
TestCfg cfg = new TestCfg(this);
cfg.load(); // loading data from config.yml
cfg.save(); // saving data to config.yml 
```

After executing cfg.save() you'll find a config.yml in your plugin folder:
```
test:
   string: "test string"
   integer: 10
   list:
   - test 1
   - test 2
   - test 3
   
```

To access to config values, you just need to use TestCfg fields as usually. After modification of the fields, you must call save() method to save data to file.
```
cfg.testInt = 100;
cfg.save();
player.sendMessage("Test string from config: "+cfg.testString);
```


Snake1999:
:+1:

Snake1999:
Good tool for config! For further work, we might do renames and deal with exceptions :D

MagicDroidX:
This is really awesome! ;)

PR #506, opened 2016-02-15
Bukkit: Implementation

PizzaCrust:
This pull request begins to implement the Bukkit API to allow users to use Bukkit plugins from the Minecraft PC world.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,6 +6,13 @@
     <artifactId>nukkit</artifactId>
     <version>1.0-SNAPSHOT</version>
 
+    <repositories>
+        <repository>
+            <id>spigot-repo</id>
+            <url>https://hub.spigotmc.org/nexus/content/repositories/snapshots/</url>
+        </repository>
+    </repositories>
+
     <properties>
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
@@ -42,6 +49,12 @@
             <artifactId>junit</artifactId>
             <version>4.12</version>
             <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.bukkit</groupId>
+            <artifactId>bukkit</artifactId>
+            <version>1.8.8-R0.1-SNAPSHOT</version>
+            <scope>provided</scope>
         </dependency>
     </dependencies>
 
--- a/src/main/java/cn/nukkit/Server.java
+++ b/src/main/java/cn/nukkit/Server.java
@@ -2,6 +2,7 @@ package cn.nukkit;
 
 import cn.nukkit.block.Block;
 import cn.nukkit.blockentity.*;
+import cn.nukkit.bukkit.BukkitServer;
 import cn.nukkit.command.*;
 import cn.nukkit.entity.Attribute;
 import cn.nukkit.entity.Entity;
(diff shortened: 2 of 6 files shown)


MrPowerGamerBR:
So Bukkit plugins won't need a lot of changes to work in Nukkit? Sweet!
Em 15 de fev de 2016 07:05, "PizzaCrust" <notifications@github.com>
escreveu:

PizzaCrust:
@MrPowerGamerBR Even better, Bukkit plugins might not need changes to work since the API acts like a abstraction layer.

Pub4Game:
It turns out to be 2 API... which can be confused... Why all this?

PizzaCrust:
To allow people to have more variety in there plugins.

MrPowerGamerBR:
As long the Bukkit API reroutes to the Bukkit API I'm fine with it, but it
would be better to be a add-on/plugin for Nukkiy, but I think that's not
possible :/
Em 15 de fev de 2016 10:27, "PizzaCrust" <notifications@github.com>
escreveu:

fromgate:
I'm agree with @MrPowerGamerBR 

Bukkit API is really great, but it must be implemented as plugin - something like bridge that allows to run some bukkit plugins.

Pub4Game:
@fromgate :+1: 

Yes, implement a few APIs in Nukkit be silly... it is Better to create an add-on (plugin) that would allow us to work with plugins from Bukkit

MrPowerGamerBR:
But... It is possible to do as a plugin?

I don't if it is, in my head you just need to create the Bukkit classes in
the same package as the Bukkit ones with all of the methods that "reroute"
to the Nukkit API, however I think it is more complex than that and I think
a plugin can't do that...
Em 15 de fev de 2016 12:20, "Pub4Game" <notifications@github.com> escreveu:

ProjectInfinity:
I'm against this. A bridge plugin should be made and can be done for those who want, but keep it out of the core.

Anyhow, it's already ridiculously easy to port to Nukkit. Adding this would serve no purpose beyond possibly causing severe negative side effects.

xxFlare:
@ProjectInfinity :+1: 

It would also add a lot of confusion.

Snake1999:
I think this should be added as a *plugin*. We'll all be happy if this plugin comes out :+1:

PizzaCrust:
Will start developing the plugin, soon.

MrPowerGamerBR:
@PizzaCrust so... how is the plugin API implementation is going? I really want to test your plugin when it is done :)

PizzaCrust:
@MrPowerGamerBR Implementing it is not hard, but there is not much people supporting me; so it's going slowly.

MrPowerGamerBR:
@PizzaCrust yeah, I tried creating my own implemention™ and the only thing I could implement was `BukkitRunnable` and `Bukkit.broadcastMessage("wow")` (Even tho I don't know how a implementation works, I only created classes and packages with the same name as the Bukkit ones, but then I got to the player class and I didn't know how I could implement it)

PizzaCrust:
The implementation is basically a API wrapper.

MrPowerGamerBR:
@PizzaCrust you could upload the code you already made (if you made something, of course) to GitHub, maybe somebody would help you ;)

mihakrut:
please release nukkit for minecraft pocket edition 0.14.0 with bukkit api :+1:

twixi:
Nedd, very need!

Tee7even:
If you need Bukkit API, go use Pokkit or something, Bukkit API or any other API (like Sponge, PM, how much you can come up with?) won't ever be a part of the vanilla Nukkit.

PR #513, opened 2016-02-17
Add PlayerChunkRequestEvent

fengberd:


ZhyMC:
What a **** event(2333)

MrPowerGamerBR:
Why do you think this is a "**** event"? I think it would be very useful if
you could edit the chunk before sending it.
Em 17 de fev de 2016 10:03, "ZhengHuaiYu" <notifications@github.com>
escreveu:

ZhyMC:
just a j oke between friends

fengberd:
Well,I add this event to cancel a chunk request,if you want to edit the chunk data ,you can use  "ServerDataPacketSendEvent"
:D

ProjectInfinity:
This should be cleaned up. Commits are a mess.

PR #541, opened 2016-02-27
More good for mini-games and setLevel returns this

MrGenga:


MrPowerGamerBR:
It would be better to create a "setUnsafeLevel" instead of just removing the limits from the "setLevel".

MrGenga:
Ok

PR #542, opened 2016-02-27
New function setUnsafeLevel

MrGenga:


src/main/java/cn/nukkit/item/enchantment/Enchantment.java
```diff
@@ -134,6 +134,11 @@ public void setLevel(int level) {
         }
     }
 
+    public Enchantment setUnsafeLevel(int level){ //WARNING: buggy
```
xpyctum: Why not give a level?

MagicDroidX:
I wonder if this will crash the client by setting a level which is out of the bound.
Or, other negative affect.

MrGenga:
No, I tested it on PocketMine, works fine.

PR #547, opened 2016-02-27
Improvement RequestChunkRadiusPacket

Pub4Game:


MagicDroidX:
Since it is the chunk radius, the view distance should be `radius << 4` i guess.

Pub4Game:
@MagicDroidX This is done in PocketMine.

MagicDroidX:
PocketMine isn't always right.
On the other hand, I have found that 
**view distance**: 
>Sets the amount of world data the server sends the client, measured in chunks in each direction of the player (radius, not diameter). It determines the server-side viewing distance. 

According to Minecraft Offical Wikipedia: http://minecraft.gamepedia.com/Server.properties.

So the chunk radius is the view distance.

Pub4Game:
Hmm... If this is incorrect, close my pull request.

PR #551, opened 2016-02-28
Fixed bug with VERY slow file saving server.log

Pub4Game:
In Java does not need a delay


PocketKiller:
You are a nukkit member, merge it.

Pub4Game:
okay :D Thanks

PR #552, opened 2016-02-28
To fix #488

### LINKED ISSUE #488: Player in Survival Mode "absorbs" the item even when the inventory is full.
Well, the title explains everything, it seems that the player picks up the item even when the inventory is full, and the picked item simply vanishes from existence.

Pub4Game:


PocketKiller:
just a question, on PHP !== and != are different, is it same on Java?

Pub4Game:
@PocketKiller In Java != same as !== or != in php

PR #556, opened 2016-02-29
Localised logger

ant3:
Moved the message localisation code to a LocalisedLogger so that it only happens in 1 place.  This makes the rest of the code cleaner.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -1242,7 +1242,7 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
                             revert = ev.isRevert();
 
                             if (revert) {
-                                this.server.getLogger().warning(this.getServer().getLanguage().translateString("nukkit.player.invalidMove", this.getName()));
+                                this.server.getLogger().warningLocal("nukkit.player.invalidMove", this.getName());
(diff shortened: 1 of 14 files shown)


ant3:
If this isn't where you want the code-base to go then just say so.  My reason for doing this is that currently everything needs to reference the Server in order to localise log messages.  This means lots of things need a reference to the server just for logging and can't be created until after the server.  This means that the server needs to create lots of things in its constructor.  If we can break this link then some things can be passed into the server and this will allow us to reduce the size of the constructor (currently ~400 lines).

ant3:
Any comments?

Pub4Game:
@ant3 
image

ant3:
It didn't have conflicts when I initially submitted it :-).  I'm quite happy to fix them, but if people are not interested in merging this change I won't bother.  I'm just looking for feedback that this pull request is OK or not (presuming I fix these new conflicts).

Snake1999:
Maybe there will be a big hurt that, people developing plugins for our software, should change their old habit and force there mind on new ones. That seems like... you know,... unbelievable.

ant3:
Sorry, I no longer have time to contribute to this project.  Good luck.

File14:
Remove this PR then?

PR #566, opened 2016-03-05
Fix tools won't be broken when tools' durability is few

yescallop:


Pub4Game:
Items and so broken...

yescallop:
@Pub4Game I can't understand what you mean

Pub4Game:
@yescallop I mean, what tools and so can be broken....

PR #574, opened 2016-03-08
Create CallbackTask.java

MrGenga:


MagicDroidX:
Use Runnable instead.

PR #580, opened 2016-03-08
I hate caves!!!

boybook:


MrPowerGamerBR:
Why remove caves?
Em 8 de mar de 2016 14:38, "boybook" <notifications@github.com> escreveu:

boybook:
Caves are too dense, makes me uncomfortable. Not remove, just be persisted in swamp biome.
screenshot_2016-03-07-01-15-32_net zhuoweizhang mcpelauncher pro

mepeisen:
And others like Caves. Some like less caves others more caves.
Suggestion for rework: such changes could be done bone by either implementing an api to inject custom biomes by plugin (so everyone is invited to customize biome generations to fit personal needs) or by configuration values.

boybook:
The map generator will be rewritten in the near future.
So because I wrote the dense caves, I have to make it better.(temporary)
Wait for the new map generator!
screenshot_2016-03-09-21-16-27_net zhuoweizhang mcpelauncher pro
I believe these caves are enough.

PR #587, opened 2016-03-11
Bugfix: If a plugin.yml does not contain permissions (empty permissions

mepeisen:
entry) there was a NullPointerException


mepeisen:
Commit 4179dd1 is only a hotfix that ensures nukkit will continue to work. However we should think about it.
1) A Task that cannot be cancelled because onCancel throws exception should be directly removed?
2) A listener that periodically throws runtime exceptions should result in deactivating the failing plugin?

PR #593, opened 2016-03-11
Blocks in creative should not end

Pub4Game:


Pentaminum:
item.setCount(item.getCount());
Isn't this a useless line?

Snake1999:
@Pentaminum No, it isn't. This line sends the data of iten count to the clients.

PR #596, opened 2016-03-13
Can customize whether the formation of the caves. fixed the chunk entity spawn bug. Added the better grass color.

boybook:
Can customize whether the formation of the caves. fixed the chunk entity spawn bug. Added the better grass color.


boybook:
I feel hopeless about it. 😭😭😱

Niall7459:
Are you not going to submit this and get it merged? Looks like a lot of hard work you have put into it.

PR #599, opened 2016-03-13
Add usable function "generationScreen"

MrGenga:
It can be used on working with Nether Portal, and more beautiful plugin's teleportations from world to world :)

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -632,9 +632,9 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
         this.generationScreen(0);
     }
 
-    public void generationScreen(byte dimension){
+    public void generationScreen(int dimension){
         ChangeDimensionPacket pk = new ChangeDimensionPacket();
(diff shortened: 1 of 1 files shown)


MrPowerGamerBR:
So... if this pull request is merged, then every teleportation (between worlds) will have the "Loading world" screen? If yes, then that would be good, better than teletransporting and thinking the server has lagged just because the teletransportation didn't happen :stuck_out_tongue: 

Also, there should be a boolean overriding the screen on .teleport, so developers can teleport without having the screen ;)

Pub4Game:
No, teleportation will take place as before.

He added functions that can be used in plugins.

boybook:
I tested it but I found I can't move after teleport!

XFuryMCPE:
Bump. Fix the unable to move bug @MrGenga

Snake1999:
This idea is good, but it will be better if we could place a trigger, that means an enable/disable switch in config, to deal with different server requirements. A "one-knife-cut" is too straight for changing it directly, instead, we should write both these functions, and add a switch :D

PR #600, opened 2016-03-13
Fixed damage calculation with armor

Pub4Game:


Pub4Game:
@MagicDroidX you check this?

PR #605, opened 2016-03-15
Small fix, omg, stupid fire

Pub4Game:


xxFlare:
This does not fix it.

In vanilla MCPE, you can still be set on fire in creative but it goes out when you step out of the fire. Nukkit needs to shorten the extinguish time instead in creative.

Pub4Game:
It works...check this

xxFlare:
Yes it works, but it's not the correct MCPE vanilla behavior we need.

> In vanilla MCPE, you can still be set on fire in creative but it goes out when you step out of the fire. Nukkit needs to shorten the extinguish time instead in creative.

Pub4Game:
It is already there https://github.com/Nukkit/Nukkit/blob/master/src/main/java/cn/nukkit/Player.java#L1409-L1417

MagicDroidX:
@xxFlare No, you can't. @pub4game  I've found that player in creative can be set on fire when in lava.

PR #607, opened 2016-03-17
Remove extra "Cancellable" in the event

Pub4Game:


MagicDroidX:
These event should be made to cancellable.
Although this is no syntax error after removing the interface, this is used by reflection when a event is  being called.

Pub4Game:
lol, PlayerQuitEvent should be cancellable?

Angelic47:
@Pub4Game  He means that these events used by reflections, if you removed this interface, java will can't find methods and may cause exception.
In my opinion, if somethings doesn't want to be cancel, you can rewrite "setCancelled()" to throw an exception when plugins called that.

Pub4Game:
@Angelic47 What exception may be in PlayerQuitEvent? What do you mean? This event is impossible to complete and no one would think to do..

fromgate:
@MagicDroidX 
So it must be properly processed where you using of reflections :)

PR #611, opened 2016-03-18
Fix attack. Now the player pushes and he gets true damage.

Pub4Game:


boybook:
Good job! This is what I need!

Pub4Game:
Thanks :3

PR #614, opened 2016-03-19
fix craft

Pub4Game:


BlackShadow1:
Are you serious? You've modified the amount to make the item.. Now if you craft bread, wheat decreases only 1 number, it must be 3, like in singleplayer...

HmHmmHm:
@MagicDroidX check please i think it is wrong

Pub4Game:
omg, X - ONE ITEM

IN CRAFT  "XXX" - "ONE ITEM, ONE ITEM, ONE ITEM"

Pub4Game:
@BlackShadow1 Check yourself before you put a negative rating...

BlackShadow1:
I don't know if it work, but in CraftingManager in PM, the amount isn't affected by the X.

Pub4Game:
Why write if you didn't check it?

BlackShadow1:
I don't know, i'll check if your method work.

PR #615, opened 2016-03-19
Added Config.load (Inputstream is)

fromgate:
Added method Config.load (Inputstream is) required to load config values (for example init, values) directly from plugin's jar file.
Example:
```
        String language = "rus";
        Config lng = new Config(language+".lng",Config.YAML);
        InputStream is = plugin.getClass().getResourceAsStream("/lang/"+language+".lng");
        lng.load(is); // load rus.lng from plugin's jar.
```


Pub4Game:
Great!

Issue #620, opened 2016-03-20
chunk loading


Creeperface01:
after few minutes chunks dont load.

ProjectInfinity:
Not getting this. Explored a map and it generated new chunks as I moved.

Issue #618, opened 2016-03-20
teleporting bug


Creeperface01:
when i teleport to unloaded or new chunks i stuck in the air. and teleport takes +/- 5 seconds but it shouln't.

MagicDroidX:
The chunks are generating I am afraid.

Creeperface01:
no they are already generated

Creeperface01:
you can test it here: play.minetox.cz

Creeperface01:
BTW why server loads around 2000 chunks before player join? o.O

Creeperface01:
and probably before teleport

Creeperface01:
i got it!

PR #622, opened 2016-03-20
Fixed chunks sending! #618 #620

Creeperface01:
- now teleporting and joining are a lot faster!
- server is faster
- you can play without horrible lags (FPS)

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -731,7 +731,7 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
         int x = 0;
         int z = 0;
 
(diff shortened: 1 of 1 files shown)


Pub4Game:
image

Constantly freezes loading maps

Creeperface01:
Really? I tested it

Creeperface01:
speed depends on nukkit.yml too

Pub4Game:
Yes((( I test this too... and world don't load :cry:

Pub4Game:
I have a standard config nukkit.yml. The world just does not boot and the game freezes

MagicDroidX:
I think I have a better way to improve it. :) I'll try to fix it tonight

Pub4Game:
And this fix is incorrect. 

Here the correct number of = is the same as a single game.

Creeperface01:
But fix it soon please

PR #631, opened 2016-03-23
Create start.sh

MrGenga:


ProjectInfinity:
Is this really necessary?

MohiPE:
Loop is really useful :)

ProjectInfinity:
@MohiPE Wouldn't you agree that the end user should make their own startup script though? Or are we supposed to provide startup scripts for every platform out there?

PR #644, opened 2016-03-29
Set nether sky API

M4Gn1T:


Snake1999:
Maybe I should add a note here that, this **directly** setting function maybe cause come problems on synchronous updating between clients and server, saving world, and something so on. APIs should be designed carefully to prevent these problems.

PR #654, opened 2016-04-01
Added creation of a portal to hell

Pub4Game:
image


Snake1999:
Wow, That's what I looked forward to for a long time. Thanks!

Pub4Game:
@Snake1999 
There is a small problem:
If you activate the portal, then all the blocks will be replaced with "NetherPortal"
The portal is not activated, if you click on the wall
The portal is not activated without blocks at the edges (those which are not included in the outline of the portal)

Snake1999:
Oh, I merged this without testing... But I can't find out how thos bug causes

Pub4Game:
@Snake1999 You can check and fix errors :)

PR #659, opened 2016-04-02
Simplify NBTInputStream and NBTOutputStream

pcbeard:
No longer subclasses FilterInputStream/FilterOutputStream, implements AutoCloseable.
This should improve performance and simplify usage since not all of
the InputStream/OutputStream API is needed.


MagicDroidX:
Any negative affect?

pcbeard:
No negative effects that I've found.

PR #666, opened 2016-04-04
Use J8 lambdas to simplify code

pcbeard:
If we're requiring Java 8, might as well use lambdas to simplify the code.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -587,12 +587,7 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
         int count = 0;
 
         List<Map.Entry<String, Integer>> entryList = new ArrayList<>(this.loadQueue.entrySet());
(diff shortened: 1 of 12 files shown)


Pub4Game:
:+1:

liach:
Change for loops to forEach calls if possible

pcbeard:
I'm not convinced that all for-loops need to be converted to forEach(). Sometimes the loop flow control is easier to understand with a break rather than a return. Also, I've received some feedback that the project owners aren't fond of lambdas in general.

onebone:
Not sure, but as I tested before, lambda is much slower than the normal code.

geNAZt:
Lambdas are much slower than normal loops since they can't get JIT optimized. This would be a huge step backwards in terms of performance.

JonathanxD:
Inner class is much slower than lambdas. http://www.oracle.com/technetwork/java/jvmls2013kuksen-2014088.pdf.

PR #668, opened 2016-04-04
Any command can override vanilla commands.

Twister915:
Also commented the method since I had trouble understanding exactly what was meant by the statement.

I would like the ability to override the /ban /msg and other vanilla commands with my plugins. I believe this would be a common use case. Any plugin that does override /ban intentionally does so, and so vanilla commands should respect that.


MrPowerGamerBR:
You can override by using some "black magic" (changing the vanilla remapping to another dummy command and then registering the command), however, I would prefer if any plugin could override vanilla commands without any workarounds :+1:

Twister915:
Any update on this?

Snake1999:
Let's see how this black magic works ;)

PR #687, opened 2016-04-10
PlayerJumpEvent

Creeperface01:


Creeperface01:
i think this would be usefull for anticheats and in that case, handling DataPacketReceiveEvent is needless :P

ProjectInfinity:
Could you give an example why you think a PlayerJumpEvent is needed? I'm not against it, but I'm not sure I see the use for it?

Creeperface01:
As i said for anticheats. Or simon says minigame :D

MrPowerGamerBR:
@ProjectInfinity it would be good for double jump plugins.

ProjectInfinity:
@Creeperface01 @MrPowerGamerBR I feel like that's too specific use cases to justify adding a whole new event.... You might as well handle packets. How many plugins could actually benefit from this? So far we're seeing minigames (which should handle packets anyhow because you often need to go outside what the API allows for) and double jump plugin.... Seriously? :|

Creeperface01:
Yes. I have kit that allows double Jump :D in my annihilation minigame

PR #691, opened 2016-04-16
Improved PlayerTeleportEvent

Pub4Game:


MagicDroidX:
The default cause is PLUGIN in CraftBukkit.

Pub4Game:
@MagicDroidX okay

PR #696, opened 2016-04-17
Fixing teleports

fromgate:
1. Player spawn/respawn/bed will not fire PlayerTeleportEvent
2. Player teleporting will fire only PlayerTeleportEvent, but will not fire EntityTeleportEvent
3. When teleport cause is null events will not fired.


Pub4Game:
Thanks!

DyirondenRu:
Good job 👍

Issue #682, opened 2016-04-09
Revert removal of Cancellable from PlayerDeathEvent 


ProjectInfinity:
As seen in https://github.com/Nukkit/Nukkit/commit/8dbf95c7a3006a081a8e8561a227abf176045625

How does this benefit us? I only see downsides, please enlighten me.

Syntaxious:
I don't see any benefits?

Pub4Game:
Please check code in PocketMine. 

https://github.com/PocketMine/PocketMine-MP/blob/master/src/pocketmine/event/player/PlayerDeathEvent.php

This is completely unnecessar

ProjectInfinity:
Amd what if you want to cancel all player deaths easily?...

ProjectInfinity:
Besides what does it matter that PocketMine does here? Cancellable is so much more useful than forcing plugin developers to use entity damage events and then check if the entity is a player and if the damage is higher than the remaining health.

This commit is a big downgrade in terms of convenience and it serves **no purpose**  as you clearly demonstrated yourself by simply pointing to PocketMine.

MrPowerGamerBR:
Does it have any side effects when you cancel the Death Event? No? Then keep the Cancellable. If it isn't causing any issues, then keep it, it is useful for plugin developers.

If we are going to say "oh but pocketminez-mp doesnt do this way!!" then we should wipe the Nukkit project and make it in PHP, because "pocketminez-mp uses X not Y"

ProjectInfinity:
@MrPowerGamerBR I'm all for keeping things similar, but not to the extent where it only has downsides. That's my point in this case. Removing Cancellable because PocketMine doesn't have it simply makes PlayerDeathEvent less convenient and doesn't improve anything.

ProjectInfinity:
@MagicDroidX Please review our concerns here.

Pub4Game:
@ProjectInfinity omg, just make pull request

ProjectInfinity:
@Pub4Game I can't tell if you're trying to act dumb on purpose or if it comes naturally. 

See it this way. I make a commit, he has to inspect it either way and accept it. Right? It requires equally much effort to just click revert on a commit and at the same time causes less garbage to be inserted into the commit history.

Pub4Game:
Of course, it is better to write a bunch of messages proving his innocence...

If you were smarter and made this pull request, the MDX has long been watched... 

You're just stalling and making it more complicated.

PR #698, opened 2016-04-17
PlayerDeathEvent should be cancellable.

ProjectInfinity:
There's no reason to not let PlayerDeathEvent be cancellable. Look at #682 for reference.

@Pub4Game Happy?


Pub4Game:
Yes, I am very happy

PR #708, opened 2016-04-23
Update EnchantInventory.java

Creeperface01:


MagicDroidX:
I cannot get your point.

Creeperface01:
I got an error undefined index -1

MagicDroidX:
That's weird. This line was added because I had got an IndexOutOfBound Exception. :/

PR #710, opened 2016-04-24
Fixed the NullPointerException when getPlayer(name) in Server class.

boybook:


boybook:
Before the server received the LoginPacket, the Player has already been added into this map! This bug will only happen when a lot of players join the server at one time.

PR #712, opened 2016-04-24
fix getOnlinePlayers()

Creeperface01:


src/main/java/cn/nukkit/Server.java
```diff
@@ -1392,7 +1392,7 @@ public Player getPlayer(String name) {
         name = name.toLowerCase();
         int delta = Integer.MAX_VALUE;
         for (Player player : this.getOnlinePlayers().values()) {
-            if (player.getName() != null && player.getName().toLowerCase().startsWith(name)) {
+            if (player.getName().toLowerCase().startsWith(name)) {
```
MagicDroidX: ？

Creeperface01:
There are not players with null name

PR #744, opened 2016-04-30
improved boat and riding

Creeperface01:


Pub4Game:
Nice :+1: I'll check it tonight

Creeperface01:
But if you place boat on ground so after reconnect it spawns in ground

Creeperface01:
And maybe every entity should have variable linkedEntity but i added it only in vehicle for now

Pub4Game:
@Creeperface01 Sorry, no time to check it out. 

Looks good, but You definitely are convinced that it works correctly?

Creeperface01:
It works better than before but death animation and boat rotation doesnt

PR #750, opened 2016-05-02
Falling - fail

Creeperface01:
Oops


Creeperface01:
or should be on ground state checked every time? because theoretically another player can place/break block under player

PR #757, opened 2016-05-04
Bb

Creeperface01:
i think air or liquid should have bounding box too

### DIFF
--- a/src/main/java/cn/nukkit/block/BlockAir.java
+++ b/src/main/java/cn/nukkit/block/BlockAir.java
@@ -58,11 +58,6 @@ public class BlockAir extends BlockTransparent {
     }
 
     @Override
-    public AxisAlignedBB getBoundingBox() {
-        return null;
-    }
-
-    @Override
     public double getHardness() {
         return 0;
     }
--- a/src/main/java/cn/nukkit/block/BlockLiquid.java
+++ b/src/main/java/cn/nukkit/block/BlockLiquid.java
@@ -41,9 +41,15 @@ public abstract class BlockLiquid extends BlockTransparent {
         return false;
     }
 
-    @Override
-    public AxisAlignedBB getBoundingBox() {
-        return null;
+    protected AxisAlignedBB recalculateBoundingBox() {
+        return new AxisAlignedBB(
+                this.x,
+                this.y,
+                this.z,
+                this.x + 1,
+                this.y + getFluidHeightPercent() - 0.1111111,
+                this.z + 1
+        );
     }
 
     @Override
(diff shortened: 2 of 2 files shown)


Tee7even:
Why?

Creeperface01:
Because player can collide with liquid or air

Tee7even:
Justify its usefulness. "Player can collide wuth them" is not an explanation. Of course, player can go THROUGH them. But there's no any use for this, because it's complete nonsense.

Creeperface01:
Player can go through more blocks that have bb too

Creeperface01:
Player can go through more blocks that have bb too

Creeperface01:
There is canPassThrough()  method for that

Tee7even:
Examples, please. And yeah, you haven't yet said why it's useful.

Creeperface01:
For collision checking?

Tee7even:
Collision with nothing?

Creeperface01:
No... Collision with air or liquid

Tee7even:
Air is literally nothing. Liquids are something likewise, they don't need a bounding box.

Creeperface01:
Liquids are important. And collision too. (in liquid player can move up) but in air not

Tee7even:
Those. Blocks. Act. Differently.

Creeperface01:
All blocks have its own bounding box and it doesnt depend on its type

Tee7even:
No, they don't, only blocks with non-standard shape have it. And it doesn't make sense to add it for nothing.

Creeperface01:
it isn't nothing, it may be nothing for players, not for developers

Tee7even:
You can't check if entity "collides" (lol) with air/liquid or what? Or do you need to do *complex computation of collision with air*?

Creeperface01:
Why not? And Do you know About plugin API in nukkit?

Tee7even:
So? There's no way to check if something "collides" with air?  Removing BBs from blocks that can be passed through makes more sense than adding them *to air and liquids*.

boybook:
The Nukkit is lag enough, please do notcmake it more lag!

Creeperface01:
With this we can improve methods like isInsideofWater / solid and make nukkit faster

Tee7even:
What will make it faster? Checking for BB collision? Sorry to dissapoint you, but it will make Nukkit slower.

Creeperface01:
You are right air desnt need bb but blocks that has some special attribute/feature like liquid, ladder, cactus need it

Creeperface01:
Method getBlock() is a lot slower than intersectsWith()

Snake1999:
No more lags and no argues here.

PR #758, opened 2016-05-06
Add bounding box for liquid

Creeperface01:
i think liquid should have BB because we check collision with liquid very often

### DIFF
--- a/src/main/java/cn/nukkit/block/BlockAir.java
+++ b/src/main/java/cn/nukkit/block/BlockAir.java
@@ -38,6 +38,11 @@ public class BlockAir extends BlockTransparent {
     }
 
     @Override
+    public AxisAlignedBB getBoundingBox() {
+        return null;
+    }
+
+    @Override
     public boolean canBeFlowedInto() {
         return true;
     }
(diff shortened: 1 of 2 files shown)


Creeperface01:
and i think we should rewrite block collision checking in nukkit

Tee7even:
As I said earlier, there's no need to add a bounding box for liquids. Collision with liquids, unlike other blocks, is mainly calculated by height.
I also would recommend you to stop doing merge commits. If you are using IDEA, then use "Rebase my GitHub fork feature", if you don't, then look here: https://help.github.com/articles/syncing-a-fork/

Creeperface01:
Yes it is!  Block collision is bad in nukkit. E. G. you dont collide with lava but nukkit sets you on fire

Creeperface01:
And as i said i try to imrpove it so it should be less laggy too

Snake1999:
Not needed.

Creeperface01:
Omg why dont you merge this? Or you thing liquid collision checking is needless??

Creeperface01:
i think you still don't understand me. block's throughput and collision are two different things.. e.g. you can go through lava or fire but you are colliding with them and getting damage.

Creeperface01:
so if you check block throughput use canPassThrough() or instanceof BlockTransparent but if you would to check collision use boundingBox

Creeperface01:
and i think performance will be still same because server doesn't check collision of blocks that haven't any effect

Creeperface01:
@Pub4Game

Pub4Game:
I'm sorry, but I'm busy, now will be the week of exams.

PR #767, opened 2016-05-11
Fixed #480

### LINKED ISSUE #480: Black grass
So black grass is a problem I can fix, but there are a couple ways we can do it.

1: Revert PocketMine's commit that caused this problem.
2: Force the color always be green.

Which way would be the best?

HmHmmHm:
If chunk block biome color is produced in black.
(or chunk block biome doesn't have any color)
LightPopulationTask is will be fixed that biome color.


Pub4Game:
It may be better to do it when loading chunks, not in LightPopulationTask

HmHmmHm:
@Pub4Game agreed, i'll change it

PR #768, opened 2016-05-11
Fixed #424

### LINKED ISSUE #424: Bug with DoublePlant
It manages to place 2 block (side(1)), as the bottom block is replaced by air

image

canBeReplaced() only Double Tallgrass

HmHmmHm:
I checked this problem,
actually Large Fern also can be replace,

Double Tallgrass
Large Fern(175:3)


Pub4Game:
image

I think it would be better

HmHmmHm:
@Pub4Game Ouch my mistake

PR #769, opened 2016-05-12
Update BlockWaterLily.java

Creeperface01:


Pub4Game:
`canPassThrough()` default **false**

Creeperface01:
Why? we can not pass through

Creeperface01:
for example you can stand on it

Pub4Game:
@Creeperface01 This will be fixed later.

Creeperface01:
ok

PR #774, opened 2016-05-17
Added fallback for plugin.yml to nukkit.yml

ConnorLinfoot:
Not sure what others think about this but personally I would like to have my PC & PE plugin all in one big file, as I currently do with Spigot/BungeeCord.


Tee7even:
Explain it in more detail, please. I didn't get it.

MrPowerGamerBR:
@Tee7even He wants to use only one JAR for a Bukkit/Nukkit plugin.

If Nukkit couldn't load the plugin by using the plugin.yml, Nukkit uses nukkit.yml to load the plugin.

BungeeCord does a similar thing, allowing users to create "Universal" plugins (One JAR that works in BungeeCord and Bukkit)

fromgate:
I think it's a great idea!!!

fromgate:
@ConnorLinfoot 
Thank you! You're great! And your idea is really awesome :)

PR #788, opened 2016-05-26
Fix's TNT destroying liquids

imjack:
Liquids kept blowing up from TNT


src/main/java/cn/nukkit/block/BlockLiquid.java
```diff
@@ -385,6 +385,11 @@ private int calculateFlowCost(Block block, int accumulatedCost, int previousDire
     public double getHardness() {
         return 100;
     }
+    
+    @Override
+    public double getResistance() {
+    	return 500;
+    };
```
liach: Extra semicolon

Pub4Game:
Thanks :3

PR #792, opened 2016-05-28
Add Lever?

boybook:
Added lever, but still some bugs.


Pub4Game:
Please remove debug lines

PR #793, opened 2016-05-28
Removal of not needed code

onebone:
BlockLiquid.java at 331 and 203, its parameter updates is set as true, which results in updating block. Thus, scheduling update is duplicating code. Also, water flows normally even if the code is removed.


KMUzinCRSG:
@MagicDroidX guys anyone can check this pull request?

PR #813, opened 2016-06-08
Player .dat performance improved

KMUzinCRSG:
When many session are coming and breakout, it will be help.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -61,10 +61,13 @@ import cn.nukkit.permission.PermissionAttachmentInfo;
 import cn.nukkit.plugin.Plugin;
 import cn.nukkit.potion.Effect;
 import cn.nukkit.potion.Potion;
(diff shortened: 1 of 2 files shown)


KMUzinCRSG:
@MagicDroidX can you check?

KMUzinCRSG:
@MagicDroidX guuuuuys check this please!
I need to work this part more- 'n'...
(ex.Player.processLogin()-> super.init(this.level.getChunk(~) )
(ex.Player.processLogin()->spawnPosition = this.getSpawn();)

but i already deleted past forked repo... :(

MagicDroidX:
I think this might cause unexpected problems when many players join and leave very frequently.

MagicDroidX:
I think this might cause unexpected problems when many players join and leave very frequently.

KMUzinCRSG:
@MagicDroidX
hmm.. agreed. i think it also need to be test,
please slow down the test reliably.
i'll do other things like a populateTask check.

Issue #818, opened 2016-06-08
Error with server.log (Too many open files)


Pub4Game:
```
[ALERT] java.io.FileNotFoundException: /home/servs/srv3610/server.log (Too many open files)
	at java.io.FileOutputStream.open0(Native Method)
	at java.io.FileOutputStream.open(FileOutputStream.java:270)
	at java.io.FileOutputStream.<init>(FileOutputStream.java:213)
	at cn.nukkit.utils.MainLogger.run(MainLogger.java:172)
```
Perhaps this is the problem of changing #659

MrPowerGamerBR:
If you are running on a Linux system, change the soft and hard file limits.
Em 8 de jun de 2016 16:24, "Pub4Game" <notifications@github.com> escreveu:

fromgate:
@KMUzinCRSG 

Me and @Pub4Game performed some tests and we found that source of these lags (Too many open files) is your modification (chunk performance improved).

I created a simple plugin that shows number of open files: https://github.com/fromgate/FileCounter/blob/master/src/main/java/ru/nukkit/filecounter/FileCounter.java

We noticed that with old nukkit builds number of open files is keep in normal values, but after your modification every logged in player added 500 open files and this number is increasing every second. After reaching OS limit - server is freezes, player files and active config files corrupts.

KMUzinCRSG:
@fromgate
Ok i'll debug it, have you tried last version?

fromgate:
I didn't - but @Pub4Game did. He tested with latest version of course. Now we will try with your latest update.

KMUzinCRSG:
@fromgate nice, well if a problem occurs after, please upload issues.

Pub4Game:
I would not want to accept it. Because the performance of my servers fell, the load is increased and "time" is always 5.0-10.0 ms

KMUzinCRSG:
I recommend make the branch.
ex. ChunkAsyncTask

Pub4Game:
First need to fix this in master.

KMUzinCRSG:
@Pub4Game I'll do that process.
can you aceept revert 2 3 pull request?

Pub4Game:
Maybe you can try to fix it?

KMUzinCRSG:
@Pub4Game Safety First :p

Pub4Game:
You will make a pull request to return source code?

KMUzinCRSG:
Yup

PR #820, opened 2016-06-09
Chunk performance improved (2)

KMUzinCRSG:
populateChunk(...) also changed asynchronously.
@MagicDroidX


KMUzinCRSG:
This patch associated with the #818

PR #821, opened 2016-06-09
Chunk performance improved (3)

KMUzinCRSG:
Changed the chunks save in an asynchronous.


Pub4Game:
Looks good, I'll accept it, but perhaps there are more errors.

Issue #828, opened 2016-06-11
setAllowFlight() should reset inAirTicks if the player is in Survival


MrPowerGamerBR:
If you use setAllowFlight in Survival and then deactivate setAllowFlight when the player is in air, the player is kicked due to "Flying is not enabled in this server!"

The inAirTicks should be set to 0 when the setAllowFlight is changed so the kick doesn't trigger.

MrPowerGamerBR:
Pull request with the fix: #829 

I'm not sure if that's the best way to handle the fix, but it does work.

PR #829, opened 2016-06-11
Reset inAirTicks when changing flight mode

MrPowerGamerBR:
Fix for #828, I'm not sure if that's the best way to implement it, if this isn't the best way, at least add a "setInAirTicks" method. :)


Pub4Game:
Please do this for` setGamemode()`

MrPowerGamerBR:
@Pub4Game done, sorry for taking so long, I didn't see your comment until now ;)

PR #833, opened 2016-06-12
Move getHandler() to Event class

Gamecrafter:
**90** files were modified in this PR.

### DIFF
--- a/src/main/java/cn/nukkit/event/Event.java
+++ b/src/main/java/cn/nukkit/event/Event.java
@@ -25,6 +25,7 @@ public abstract class Event {
 
     protected String eventName = null;
     private boolean isCancelled = false;
+    protected static HandlerList handlers = null;
 
     final public String getEventName() {
         return eventName == null ? getClass().getName() : eventName;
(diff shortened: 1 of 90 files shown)


Pub4Game:
I think it should be _**public**_
```protected static HandlerList handlers = null;```

Gamecrafter:
@Pub4Game I don't know, then. Let's wait for more people's words on this.

Pub4Game:
Just check pocketmine - https://github.com/PocketMine/PocketMine-MP/blob/master/src/pocketmine/event/Event.php#L32

Gamecrafter:
That's only when a sub-class extends off of it.

Pub4Game:
Hmm... yeah...

image

Pub4Game:
image

Gamecrafter:
What class did you get that from?

Pub4Game:
In this case _PlayerEvent_

Gamecrafter:
There is no `getHandlers()` method in PlayerEvent. Are you sure it's PlayerEvent?

Pub4Game:
Why not just made a simple plugin for testing Eventa? For example, change the motd, then you'll see for yourself.

Gamecrafter:
Merge it, and I'll (or you if you want to) fix it in another commit, yeah?

PR #836, opened 2016-06-12
Move getHandlers() to Event class

Gamecrafter:
**90** files were modified in this PR. It should work this time.

### DIFF
--- a/src/main/java/cn/nukkit/event/Event.java
+++ b/src/main/java/cn/nukkit/event/Event.java
@@ -25,6 +25,7 @@ public abstract class Event {
 
     protected String eventName = null;
     private boolean isCancelled = false;
+    public static HandlerList handlers = null;
 
     final public String getEventName() {
         return eventName == null ? getClass().getName() : eventName;
@@ -48,5 +49,10 @@ public abstract class Event {
         isCancelled = value;
     }
 
-
+    public HandlerList getHandlers(){
+        if(handlers == null){
+            handlers = new HandlerList();
+        }
+        return handlers;
+    }
 }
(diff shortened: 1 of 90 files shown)


Gamecrafter:
@MagicDroidX

Tee7even:
Just... don't. There's so much wrong with it, so I don't know where to start.

First of all, you know, there must be a reason why those 357 lines of code are still there and not replaced by shorter alternative. And let me introduce you to it. The reason behind this is that not all events have handler lists. Events that are extending other events don't. And they can't be handled with their own handlers. Instead you declare a handler for parent event and handle it somehow like this:
```
//pseudocode
if(entityDamageEvent instanceof EntityDamageByEntityEvent) {
}
```
I am, myself, not a fan of such style, but it's an **API change**, it's serious and it's better not to touch it, until someone, who is responsible, says their word.

If the change is to be made anyway, you should think it over really carefully. Such a change is not made just by removing `HandlerList`s from events and adding one to... Oh, wait you didn't even remove them, lol, you just removed the methods... Anyway, you got the point, you have to go through the whole event subsystem of the server to know if your change affects something. So if you are going to do it in my opinion you have to ensure that the old style of event handling still works and nothing is broken by testing the sheet out of your commit (actually, that should be done to any commit, but at least do it with such changes).

All right, let's move on. Why `handlers` is public? Why having a getter method for public field?

And at last, but not least, what a wonderful, helpful and meaningful commit message!

Pub4Game:
@Tee7even He did, because I made a mistake and said it... I'll try to fix it today.

Tee7even:
Fix what?
14 июня 2016 г. 15:27 пользователь "Pub4Game" <notifications@github.com>
написал:

Pub4Game:
Need to do getHandlers() is required.

Pub4Game:
Read more to Slack. Come, discuss.

PR #841, opened 2016-06-15
Update temperature and rainfall calculation

Niall7459:
Not sure if equivalent numbers?


Niall7459:
Not sure what difference it will make /:

Niall7459:
Wrong branch

PR #856, opened 2016-06-20
Fix wrong slots number

yescallop:
The slots number in 0.15.0 has been changed to 45


Pub4Game:
I think this fix is wrong.
Inventory - 36
Hot bar - 9
Armor - 4

36 + 9 + 4 = 49

yescallop:
@Pub4Game But the inventory slots in Nukkit is 27, It seems 0.15.0 add some slots, not inventory

PR #866, opened 2016-06-22
RCON

Tee7even:
**Do not merge yet.**

Now I present you something that started as port of PM's RCON implementation, but then switched to more Java(-ish) way.  _After nine years in development, hopefully, it will have been worth the wait._
Despite the fact that I have tested it a bit for some common use cases, I ask **you** to test it, I don't trust myself. ¯\_(ツ)_/¯ So test it with whatever client that you think works best. You can find a testing build [here](https://github.com/Tee7even/Nukkit/releases/tag/test-build). Report error messages here, if you see any.


yescallop:
I have tested it, works well

Tee7even:
Bug(s) found: multiple clients issuing two requests per second, when they disconnect, this may happen:
image
This, I guess, happens because of not cleaning send queue, when client disconnects. But, if done so, then happens this:
image

_Any help appreciated._

Tee7even:
@MagicDroidX Do you even read before merging? It would be better to first fix it and then merge it.

PR #891, opened 2016-06-26
Fix teleport

MrGenga:
Revert old "Fix teleport", because it doesn't teleport to other worlds.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -3729,80 +3729,191 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


Pub4Game:
Teleportation to different worlds works fine.

Example:
```
if (!sender.getServer().isLevelGenerated(args[0])) { // World is not found
   return false;
} else if (!sender.getServer().isLevelLoaded(args[0])) { // World is not loaded
   // Loading world
   if (!sender.getServer().loadLevel(args[0])) { // Error loading
      return false;
    }
}
((Player) sender).teleport(this.getServer().getLevelByName(args[0]).getSafeSpawn()); // Teleport to world
```

fromgate:
#blamevectors

:)

PR #893, opened 2016-06-26
Update Player.java

Creeperface01:


Pub4Game:
New message?

Creeperface01:
No.. I think server shouldnt kick player that is already in game

Pub4Game:
if `kick()` is **false** then _this.close()_ it normal

Creeperface01:
... But how to check why player got kicked.... Should we check message? It is bad...

Creeperface01:
And without plugin is not event cancelled. And tell me why should server kick player that is in game -,- fuck the logic

Pub4Game:
image

Creeperface01:
Hmm? Maybe I dont understand why

fromgate:
> And without plugin is not event cancelled. And tell me why should server kick player that is in game -,- fuck the logic

I think by default - to prevent "freezed player". For example your connections is lost and you trying to reconnect. Now "old player" will kicked.

Tee7even:
@fromgate If connection is lost, player will disconnect after some time. I think, it's just a PM copy or Minecraft default server's behaviour. Anyway, this can be changed, but I'm not sure that it's implemented right in this PR.

fromgate:
@Tee7even 
But it could take a long time for w10 clients (I don't know why :))

And I think calling kick method is required to inform plugins that player will kicked now. 

So this implementations is OK (But I think messages must be localized).

Creeperface01:
Ok if player would have same UUID we can disconnect already playing one

Tee7even:
@fromgate 
>And I think calling kick method is required to inform plugins that player will kicked now.

Did you really check it or just think so?

fromgate:
@Tee7even 
https://github.com/Nukkit/Nukkit/blob/master/src/main/java/cn/nukkit/Player.java#L3079-L3104

@Creeperface01 
But what if you lost connection on your phone and trying to connect using w10 client? UUID will not be equal, but old client could be freezed.

Anyway, using "kick" method make this part of code controllable by plugin.
 If kick event was cancelled - old player will stay on server, but new player will kicked out.

Pub4Game:
```
        for (Player p : new ArrayList<>(this.server.getOnlinePlayers().values())) {
            if (p != this && p.getName() != null && this.getName() != null && Objects.equals(p.getName().toLowerCase(), this.getName().toLowerCase())) {
                if (!p.kick("logged in from another location")) {
                    this.close(this.getLeaveMessage(), "Logged in from another location");

                    return;
                }
            } else if (p.loggedIn && this.getUniqueId().equals(p.getUniqueId())) {
                if (!p.kick("logged in from another location")) {
                    this.close(this.getLeaveMessage(), "Logged in from another location");

                    return;
                }
            }
        }
```

p - Player on server
this - Player which trying to connect to the server

Pub4Game:
@fromgate Ohh... Yes :+1:  Thanks

Creeperface01:
Player disconnects in 10 seconds so sooner than player launch win 10 edition and connects

Creeperface01:
And this is very exceptional case

Tee7even:
>@Tee7even 
https://github.com/Nukkit/Nukkit/blob/master/src/main/java/cn/nukkit/Player.java#L3079-L3104
What's the problem anyway, if, when player disconnects, PlayerQuitEvent is called?
https://github.com/Nukkit/Nukkit/blob/master/src/main/java/cn/nukkit/Player.java#L3202

Tee7even:
>Anyway, using "kick" method make this part of code controllable by plugin.
If kick event was cancelled - old player will stay on server, but new player will kicked out.

@fromgate There's also PlayerPreLoginEvent, which is called [just before](https://github.com/Nukkit/Nukkit/blob/master/src/main/java/cn/nukkit/Player.java#L1831) this, so plugins can implement this kick themselves if this is **really** needed.

Creeperface01:
So you said: "you can not use nukkit without plugins"

Creeperface01:
Im sure if you make some poll so more players will have same opinion like me

Creeperface01:
It should kick joining player by default

Creeperface01:
Anyway you should remove getName() != null

Creeperface01:
Wait omg :D sorry

Creeperface01:
OMG :D Sorry :D

PR #894, opened 2016-06-26
fix memory leak

mhsjlw:
courtesy of @dktapps

explanation:

**Before patch**:
Chunk is unloaded and saved. All entities on the chunk are saved. Level::removeEntity is called on all entities in that chunk to remove them.
Level::removeEntity kills non-player entities. Result is items and XP being spawned in an unloaded chunk.
Movement updates to these entitities (falling checks etc.) trigger requests to the blocks around the entities. Because the entities are in unloaded chunks, these requests cause the chunks to be reloaded. When the chunk is reloaded, the saved entities are spawned again.
Process repeats, leaking more and more entities until memory limit is exhausted.

**After patch**:
Chunk is unloaded and saved. All entities on the chunk are saved. Level::removeEntity is called on all entities in that chunk to remove them.
Because Level::removeEntity now uses close() instead of kill(), no items or exp are dropped, which prevents the leak from beginning.

### DIFF
--- a/src/main/java/cn/nukkit/level/Level.java
+++ b/src/main/java/cn/nukkit/level/Level.java
@@ -2232,7 +2232,7 @@ public class Level implements ChunkManager, Metadatable {
             this.players.remove(entity.getId());
             this.checkSleep();
         } else {
-            entity.kill();
+            entity.close();
         }
 
         this.entities.remove(entity.getId());


fromgate:
Omg! What a long description and short fix code! :) 
I hope it will help!

Great job :)

👍

Tee7even:
Sounds promising, but needs some checking.

mhsjlw:
also see https://github.com/iTXTech/Genisys/issues/194

dktapps:
Spread the news far and wide. Everyone deserves this fix. I spent a LOT of time fixing this.

MrPowerGamerBR:
I agree with @fromgate, you wrote a loooong description for just one line change :stuck_out_tongue: (no, it isn't a bad thing, maybe if you didn't wrote that description, people would just ignore the pull request because they think that it isn't a issue)

I can't wait until this is pulled into the master branch :smile:

dktapps:
For anyone who didn't spot it, this fix is courtesy of @dktapps a.k.a. myself. The original Genisys commit can be seen here: https://github.com/iTXTech/Genisys/commit/e98738ca2d104ecc9e38436949c6b0854db20474

Tee7even:
When chunk is unloaded, actually all entities are `close`d. So this doesn't actually do anything. Except for adding some confusion.

dktapps:
Seems your chunk loading/unloading system is very slightly different to PM and forks. In PocketMine, `removeEntity()` is called instead of `close()`. So you're right. @mhsjlw , you may as well close this.

Tee7even:
@dktapps Aaaand also, did you take into consideration that `close` calls `removeEntity` and it now calls `close`? Could you, please, comment on this?

dktapps:
@Tee7even In PM this isn't an issue, because when you call `close()` once, `closed` is set to `false`. `close()` checks if the entity is already closed. If it is, it simply returns. If it didn't then you'd end up with a problem.

I've just checked Nukkit's code and yours does exactly the same. Therefore this would not be a problem regardless.

dktapps:
Actually I tell a lie. There _is_ no difference between our chunk loading systems. @mhsjlw we were right XD

@Tee7even you just outwitted yourself and proved us right.

dktapps:
@tee7even The whole problem is close() calling removeEntity() calling kill(), causing items to drop and starting the big messy cycle above.

dktapps:
I was half-asleep when I wrote those comments above and didn't check the actual code first.

Tee7even:
>The whole problem is close() calling removeEntity() calling kill(), causing items to drop and starting the big messy cycle above.

This doesn't look to make too much difference.   Even if it does, you should take all associated stuff in consideration. Anyway, I'll going to check it again some time in the future, if before that it's going to be approved in @PocketMine, **then** we can consider it as right fix.

dktapps:
@Tee7even actually, it does. **I tested this heavily myself.** It took me weeks to trace it back to this. If you're still having problems with Genisys then this is not the cause.

dktapps:
I have a friend whose server was running Genisys and would crash every 4 hours or so due to thousands of leaked entities. It's currently been running 24 hours at a stable 40MB **since updating with the fix**, and not a single leaked entity.

Ever heard of chaos theory? One tiny change can have huge effects? This is a good example. What you say doesn't _look_ to make much difference actually has huge effects.

Pub4Game:
@dktapps I see no reason to do `close()` because `this.close` will be `true `and the function will not work.

With the same success it was possible to make:

```
public void removeEntity(Entity entity) {
    if (entity.getLevel() != this) {
        throw new LevelException("Invalid Entity level");
    }

    if (entity instanceof Player) {
        this.players.remove(entity.getId());
        this.checkSleep();
    }

    this.entities.remove(entity.getId());
    this.updateEntities.remove(entity.getId());
}
```

File14:
Add it.

xxFlare:
@MagicDroidX Did you test this?

Pub4Game:
@MagicDroidX Ohh.... No

xxFlare:
@Pub4Game I just started using Nukkit from PocketMine, but is their a way to disable server.log and where do I set the memory limit?

File14:
In your start.bat or start.sh file. @xxFlare

xxFlare:
@File14 Uh, I do 'java -jar Nukkit.jar', and dont have start.sh or start.bat.

MagicDroidX:
@Pub4Game I've checked the code and this fix is justified. What are you worrying about?

File14:
java -Xmx2G -Xms2G  -jar Nukkit.jar @xxFlare

MagicDroidX:
Detailed explaination of `-Xmx` and `-Xms` can be found here:
https://docs.oracle.com/cd/E13150_01/jrockit_jvm/jrockit/jrdocs/refman/optionX.html

PR #895, opened 2016-06-26
Fixed typo

Creeperface01:


Pub4Game:
What? Why is that?

Creeperface01:
What is difference between dataPacket() and directDataPacket() ?

Creeperface01:
@Pub4Game hey merge it please.. i need some branch compatibile with nukkit's

Pub4Game:
I see no arguments why it code should be removed

Creeperface01:
as i said what is difference between dataPacket() and directDataPacket() ?

Pub4Game:
dataPacket() - **FALSE**  https://github.com/Nukkit/Nukkit/blob/master/src/main/java/cn/nukkit/Player.java#L813

```
Integer identifier = this.interfaz.putPacket(this, packet, needACK, false);
```

directDataPacket() - **TRUE** https://github.com/Nukkit/Nukkit/blob/master/src/main/java/cn/nukkit/Player.java#L843
```
Integer identifier = this.interfaz.putPacket(this, packet, needACK, true);
```

Creeperface01:
hm but now we can not send packets from another threads, so it should be without event

Pub4Game:
You also claimed to have done everything right. Maybe it's worth to test your changes before doing a pull request? https://github.com/Nukkit/Nukkit/pull/893

Creeperface01:
yes sorry. So what do you mean about this?

Pub4Game:
We'll discuss it

fromgate:
> yes sorry, but now we can not send packets from another threads, so it should be without event

But what if you need to handle this event and going to cancel it?

Creeperface01:
it's like teleportImmediate()

PR #908, opened 2016-06-30
Update BlockTransparent.java

Creeperface01:


Creeperface01:
@Pub4Game why don't you merge this? :D

Pub4Game:
Just because it's wrong, for example the anvil, ladder, etc... is transparent and solid blocks

Creeperface01:
so then something is wrong. So if block is transparent and solid so it should extend BlockTRansparent and implement method isSolid(){ return true;} ?

Pub4Game:
Why complicate things?

Transparent solid blocks much more what transparent and not solid.
In any case, this is not correct, because you are ruining other blocks, where not specified isSolid() is true

Creeperface01:
yes i undestand but transparent solid blocks should extend Transparent class and implement method isSolid() or extend Solid class and implement method isTransparent() :D

PR #911, opened 2016-07-01
Fixing players movement

Creeperface01:


Creeperface01:
So probably not :/

fromgate:
Did it really help? Why close pull? It was not tested yet

Creeperface01:
i think it is same like MovePlayerPacket

PR #913, opened 2016-07-01
Fix Still Water doesn't flow around

yescallop:


yescallop:
Why not merge this, is this wrong fix? :)
@MagicDroidX  @Snake1999  @Pub4Game

Pub4Game:
Just no time to check :)

yescallop:
@Pub4Game OK :)

PR #925, opened 2016-07-08
Revert "More compact setLevel" because it very buggy

MrGenga:


Creeperface01:
why?

File14:
It's good that it got reverted, because now my code with ItemStack for enchantments is fully working again! ;D

PR #930, opened 2016-07-08
Fix movement

Creeperface01:


Pub4Game:
I think this could be implemented better. A little aside this pull request

src/main/java/cn/nukkit/utils/ConfigSection.java
```diff
@@ -136,17 +136,17 @@ public ConfigSection getSection(String key) {
...
+     * c1:
+     * c2:
+     * a2:
+     * b2:
+     * c3:
+     * c4:
+     * a3: true
+     * a4: "hello"
+     * a5: 100
      * <p>
```
fromgate: Can you rollback this part? Here was example of config file.

Pub4Game:
@Creeperface01 check it :)

https://github.com/Pub4Game/Nukkit/commit/ad7d521c7ec69c5a56d83281763298913d670f06

PR #932, opened 2016-07-10
Add network latency measurement

Tee7even:
Now you can get ping value for every player.


Tee7even:
@Pub4Game what's the matter, mate? 😕

PR #940, opened 2016-07-16
Fixed MTU size

onebone:
https://github.com/Hydreon/Steadfast2/commit/06b7252ec3433e89205af7b72c5d49dc8b742d52


Pub4Game:
Did you check it or just copied code?

onebone:
I have copied the code, but it seems right.

Pub4Game:
@MagicDroidX ?

MagicDroidX:
In my memory, the MTU sent by client is 548 and 1464.

caoli5288:
Did it be able to fix any bug or bring any improvements?

Snake1999:
This PR requires a deeper discussion that, how microsoft and mojang make their softwares. As for now, this number will not change as the client behaves correctly. 
We should need to have this discussion if the size is changed by client (hope it will not happen ) :D

PR #961, opened 2016-08-01
Timings v2

Tee7even:
It's here. ( ͡° ͜ʖ ͡°) Guess, I killed the whole month for this... Erhm... But at least it's finally here.
Of course, it will need some correction (especially about how timings are actually called) and Nukkit sends reports as Sponge for now, [BUT JUST LOOK AT IT](https://timings.aikar.co/v2/?id=dfd7262828f94881825daa056b5d1663).

Corresponding language update: https://github.com/Nukkit/Languages/pull/37


File14:
Wow, awesome work!

Pub4Game:
@MagicDroidX Update languages... Only that prevents us from accepting this wonderful pull request.

PR #964, opened 2016-08-04
Names with spaces.

Tee7even:
And changes in command processing related to them. Escaping of spaces, escaping of quotes, escaping of escaping... Do we need to go deeper?

A small tutorial:
Space can be escaped like this - `this\ is\ the\ same\ argument`
and like this - `"this is one argument too"`,
you can escape the quotes like this - `\"`,
backslash itself can be escaped too - `\\`.

### DIFF
--- a/src/main/java/cn/nukkit/command/SimpleCommandMap.java
+++ b/src/main/java/cn/nukkit/command/SimpleCommandMap.java
@@ -163,7 +163,7 @@ public class SimpleCommandMap implements CommandMap {
 
             if (sb.charAt(i) == ' ' && notQuoted) {
                 String arg = sb.substring(start, i);
-                if(!arg.isEmpty()){
+                if(!arg.isEmpty()) {
                     args.add(arg);
                 }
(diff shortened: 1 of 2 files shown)


src/main/java/cn/nukkit/command/SimpleCommandMap.java
```diff
@@ -141,9 +141,49 @@ private boolean registerAlias(Command command, boolean isAlias, String fallbackP
...
+                start = i + 1;
+            } else if (sb.charAt(i) == '"') {
+                sb.deleteCharAt(i);
+                --i;
+                notQuoted = !notQuoted;
+            }
+        }
+
+        String arg = sb.substring(start);
+        if(!arg.isEmpty()) {
```
Pub4Game: :angry:

Tee7even:
@Pub4Game E N J O Y

Pub4Game:
Oooo yeah.... Excellent formatting

Issue #745, opened 2016-05-01
Enchantment items in chest


DyirondenRu:
When I put enchantement items to chest and restart the server - items will be lost enchantment.

Before restart:
![screenshot_2016-05-01-10-31-30](https://cloud.githubusercontent.com/assets/18498646/14940107/4a501b48-0f88-11e6-9784-3b1ec285484b.png)

After restart:
![screenshot_2016-05-01-10-32-59](https://cloud.githubusercontent.com/assets/18498646/14940109/6252e0f4-0f88-11e6-8771-8996a968da6a.png)

wasted123:
That's a bug

PR #968, opened 2016-08-05
Enchanted items in chests save on server restart

imjack:
It works although I'm not entirely sure the Item needs to be cloned before adding the enchants. Issue #745

### DIFF
--- a/src/main/java/cn/nukkit/blockentity/BlockEntityChest.java
+++ b/src/main/java/cn/nukkit/blockentity/BlockEntityChest.java
@@ -9,6 +9,7 @@ import cn.nukkit.inventory.DoubleChestInventory;
 import cn.nukkit.inventory.InventoryHolder;
 import cn.nukkit.item.Item;
 import cn.nukkit.item.ItemBlock;
+import cn.nukkit.item.enchantment.Enchantment;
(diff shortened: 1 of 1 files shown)


Pub4Game:
This fix looks obviously wrong.

imjack:
Although you say it looks obviously wrong do you mind explaining how its wrong? Or at least a better method of fixing?

Pub4Game:
Well, you re-enchant an item, when you put or take the item...

Another solution I can't give because I didn't know the cause

imjack:
setItem is called in saveNBT otherwise how would the items enchant data be saved. I don't  think you would use that method to actually set items anyway since you'd get the inventory from the chest then set them. In the BlockEntityChest constructor getItem is called to set the item in the chest inventory which is getting inventory data and actually setting it.

Creeperface01:
I'll try it make better

Creeperface01:
@981

Creeperface01:
#981

MagicDroidX:
This should be fixed by commit: https://github.com/Nukkit/Nukkit/commit/4c3f18729fd9bc9f481ea2ed894b848997afcf25
Please close this issue after you have tested it. :)

PR #978, opened 2016-08-09
Fix speedmine detection

Creeperface01:


TheDiamondYT1:
Why do you never just merge it, instead of sqash and merge

Issue #466, opened 2016-02-06
[Feature] API Update, but which way? Share your idea!


Snake1999:
We renamed internal structures these days, everyone are talking about it. I first think why we discussed it, and finally I know is the problem about Nukkit API. We sometimes compare with bukkit:
Minecraft Kernel + Bukkit API = CraftBukkit
Nukkit Kernel + Nukkit API = Nukkit
And I see, the lack of API, can confuse developers, and add many inconveniences. We think it must be added that Nukkit should have a set of API, or Nukkit won't be Nukkit.

We should add APIs. How to add API is easy, but figuring out **which way** need everyone to discuss. We can start from: Put into which package? Have a structure like what? 

You are not just talking. You are designing the API structure, you are designing the future of Nukkit, and the way you works on it. Have fun sharing idea, and we will make all API available **just the way you like and need to**.

Welcome to share your idea! :D

Snake1999:
ok I've got it, add both bukkit and pocketmine api adapt. Add package cn.nukkit.material. :D

Pub4Game:
:smile:

PeratX:
@Snake1999 So cold here...

fromgate:
I would like to suggest an additional configuration class: https://github.com/Nukkit/Nukkit/pull/486
This class is not breaking original Config, but just adds ability to work with config files easier.
So old plugins will work too and plugin developers could chose the how to work with config.

I got the idea for SimpleConfig from popular bukkit plugin  [Yamler](https://www.spigotmc.org/resources/yamler.315/) (It was recreated for Nukkit too - [Yamler for Nukkit](http://nukkit.ru/resources/yamler.37/).
But I think this idea is really good and must be part of Nukkit, instead of external plugin (or library).

It's a minor API addition, but I think it would be really useful.

SkygridRex:
in my opinion I agree that the Nukkit API should be restructured similarly Bukkit server software to facilitate the porting of Bukkit plugins to Nukkit, it would be a matter of just rename file paths. Think about it

MrPowerGamerBR:
@SkygridRex but now porting PocketMine plugins to Nukkit is easier because the Nukkit API is almost the same API from PocketMine.

SkygridRex:
@MrPowerGamerBR  But... Bukkit hace better plugins

swift2plunder:
[...]

PR #987, opened 2016-08-12
[ATTENTION] TextFormat - this will break your plugin!

Pub4Game:
I think there too much code in Nukkit that was directly translated from PM without using Java's alternatives. So I decided to make this a pull request that changes the TextFormat class. 

### Now the TextFormat is not a String, so you will need to recompile all your plugins that use this class.
image

I hope you will understand that these changes are necessary :smiling_imp: 

+1 https://github.com/Nukkit/Nukkit/issues/466

**NOT TO ACCEPT THIS PULL REQUEST**


File14:
What's the point?...

Tee7even:
@File14 API enhancement.

Issue #963, opened 2016-08-02
Enchant Table Not Working...


yungtechboy1:
### Issue description

For some reason when I try and Enchant the Item does not go in my Inventory... I did some debuging and it seems that it is the the Enchant Inventory but is never added to or sent to the player's inventory so when you try and put that item in your hand the server says you dont have that item... I would Debug it my self but compiling the Nukkit Project was a bit too much for me to do right now
### Steps to reproduce the issue

Just try and Enchant an Item then add it to your Inv
### OS and versions

Newest Build...

Creeperface01:
enchanting is not fully implemented yet

shanwer:
We didn't write enchant table function yet,but you can use /enchant command to enchant your item.

Creeperface01:
wait, enchant should work i just checked source code hmm....

yungtechboy1:
Yea That what I fixed in my Push

Snake1999:
Can you try if you can enchant again in the latest build? @yungtechboy1

yungtechboy1:
@Snake1999 Sure Ill Check Now :)

PR #993, opened 2016-08-14
Fixed Enchating (#963)

yungtechboy1:
It was pretty simple to fix... It was looking for Blue which is 11 when it should have been looking for Yellow which is 4.
Since in MCPE 4 Is the Metadata for Lapis Dye.

I updated the Experience feature because you would get an error sometimes.


Creeperface01:
maybe you should fix this in Dye class instead

yungtechboy1:
I did but they denied it last time

PR #1011, opened 2016-08-29
Better getLineOfSight and getTargetBlock

yescallop:
Now uses Integer[] instead of strange Map


Snake1999:
Great!

File14:
Looks good! Can I use getBlock().getState() now? :)

minecrafter:
There's some very significant issues with this code (that #1012 doesn't fix):

* Binary searches assume that the array or list provided has already been sorted. This code doesn't sort the array provided, so results are undefined.
* This uses a boxed primitive array when `int[]` would have been better-suited (and more efficient).

PR #1021, opened 2016-09-01
better block breaking

Creeperface01:
Yes it is necessary, because break time is never accurate

### DIFF
--- a/src/main/java/cn/nukkit/level/Level.java
+++ b/src/main/java/cn/nukkit/level/Level.java
@@ -1551,7 +1551,7 @@ public class Level implements ChunkManager, Metadatable {
                 breakTime *= 1 - (0.3 * eff.getLevel());
             }
 
-            //breakTime -= 0.1;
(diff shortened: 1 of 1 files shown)


Creeperface01:
maybe -0.15 but 0.1 should be sufficient

Creeperface01:
and what about lastBreak? it should be updated evend if even is canceled shouldn't it?

MagicDroidX:
Please provide the evidence of this change.

Creeperface01:
just try to break different blocks with appropriate tool

Snake1999:
I didn't find out where the PR improves. We will discuss this if something else is find out to be correct.

Creeperface01:
its like block collision checking... try to check player block collision and revert it when player collides with solid blocks.... without reducing player bounds its not possible... and its the same with block breaking.. time that we calculate is just average breaking time so it can takes more or less time, but problem is only if its higher than average time...

PR #1023, opened 2016-09-01
Insignificant changes for TextFormat

Tee7even:
Adding some common sense


Snake1999:
Fantastic!))

fromgate:
Works fine! 120% faster!!

Tee7even:
@fromgate more irony pls. It could be left unchanged, of course, but should it?

Pub4Game:
@Tee7even Why not? I missed it... Thanks for the correction

fromgate:
@Tee7even Sometimes jokes are just jokes even if there's no smiley after them

PR #1029, opened 2016-09-08
SimpleCommand API

Tee7even:
Allows simple command handling, when there's much commands (compared to both creating a separate class for every command and switching them in "onCommand" method). =)

## Usage
```
public class SomeCommandsClass { //Can be any class, but it makes sense to create a class just for commands :)
    @Command( //This annotation is necessary to register a command
            name = "test", //this annotation field is necessary
            description = "Just a SimpleCommand test",
            usageMessage = "/test <> []",
            aliases = {"t", "tst"}
    )
    @Arguments(min = 1, max = 2) //Optional, you can specify either maximum or minimum number of arguments
    @CommandPermission("simplecommand.test") //Optional
    @ForbidConsole //Optional
    public boolean test(CommandSender sender, String label, String[] args) {
        sender.sendMessage("Hello, sender! Your args:");
        for (String arg : args) {
            sender.sendMessage(arg);
        }
        return true; //If false is returned, sender will get a usage message
    }
}
```
To register commands in the object, call to `registerSimpleCommands` in, for example, `onEnable` method of your plugin:
```
    @Override
    public void onEnable(){
        getServer().getCommandMap().registerSimpleCommands(new SomeCommandsClass());
    }
```

### DIFF
--- a/src/main/java/cn/nukkit/command/CommandMap.java
+++ b/src/main/java/cn/nukkit/command/CommandMap.java
@@ -14,6 +14,8 @@ public interface CommandMap {
 
     boolean register(String fallbackPrefix, Command command, String label);
 
+    void registerSimpleCommands(Object object);
+
     boolean dispatch(CommandSender sender, String cmdLine);
 
     void clearCommands();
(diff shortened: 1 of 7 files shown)


Tee7even:
Oh, completely forgot about `CommandPermission` :P

fromgate:
Great!
I think it will be really helpful!

Snake1999:
Awesome )

PR #1033, opened 2016-09-16
fix windows 10 crafting

Creeperface01:


Snake1999:
Lets see if this works or not :D

animator013:
When I run the server from current master I get nullpointer exception.

Also here https://github.com/Nukkit/Nukkit/blob/092f1b9de64a306096f3e7d2c3f100eed978ac0e/src/main/java/cn/nukkit/Player.java#L2763. You are trying to put something to null ArrayList.

File14:
Is it working or not?

pesoklp13:
Nope it is not working...

File14:
D:

PR #1041, opened 2016-09-18
InventoryType

Creeperface01:


Creeperface01:
OK idk why there is my already merged commit but i think it doesn't matter

Snake1999:
Should I merge this? In my opinion, I think this might be an API-breaking update. 
Take your consideration. ))

File14:
xD

PR #1054, opened 2016-09-20
Add anvil

Creeperface01:
renaming works well, but if player want to combine 2 items without changing name result item won't appear in 3rd slot... it appears only if player changes a name, so should we add item there?

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -93,6 +93,11 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
     public static final int SURVIVAL_SLOTS = 36;
     public static final int CREATIVE_SLOTS = 112;
 
+    public static final int CRAFTING_SMALL = 0;
+    public static final int CRAFTING_BIG = 1;
+    public static final int CRAFTING_ANVIL = 2;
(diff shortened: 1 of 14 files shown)


File14:
Add ;DDDD

Pub4Game:
@File14 Enough to flood, it does not speed up the process, but so frustrating

Creeperface01:
Is it OK?

Snake1999:
Oh, conflicts 
Please use `git fetch` in your local git bash, and add a commit. Thank you :)

BTW, this is an awesome update :)

NycuRO:
@Creeperface01 Max level of enchantments is 5. 

Use wiki : http://minecraft.gamepedia.com/Enchanting

NycuRO:
@Creeperface01 You can create one method : 
each enchant declare that his maximum level is x . Ie not generally 4 or 5. Each enchant to have x max by a method .

Creeperface01:
You can not get sh 5 in enchant table... Only in anvil

Creeperface01:
@Snake1999 thanks :)  i recreate this today together with more changes

NycuRO:
@Creeperface01 Ok creeper .thanks :)

Creeperface01:
#1056

PR #1059, opened 2016-09-24
Update InventoryType.java

Creeperface01:
let's see if it works


NycuRO:
@MagicDroidX Is correct this.

NycuRO:
@MagicDroidX With this you can fix dropping with other slot.

NycuRO:
https://github.com/iTXTech/Genisys/issues/1961

PR #1069, opened 2016-10-03
Some fixes / optimizations

boy0001:
Fixes Nukkit not handling any packets
Use Long/BlockVector3 instead of strings for maps
Reduce cache size significantly by not storing redundant data
Reduce login time significantly by reducing min chunk radius
Reduce chunk load by avoiding processing extra data
Reduce memory usage by storing cache in weak map
Reduce memory usage by unloading chunks more often
Reduce memory usage by using soft reference for some block updates
- If gc'd it will send a chunk update instead (less memory usage)

Optimize collision checks
Optimize various things by reusing objects and avoiding unnecesary
creation
Add debug level 0 = does not log to file

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -42,7 +42,6 @@ import cn.nukkit.level.ChunkLoader;
 import cn.nukkit.level.Level;
 import cn.nukkit.level.Location;
 import cn.nukkit.level.Position;
-import cn.nukkit.level.format.Chunk;
 import cn.nukkit.level.format.FullChunk;
 import cn.nukkit.level.format.generic.BaseFullChunk;
 import cn.nukkit.level.particle.CriticalParticle;
@@ -70,7 +69,6 @@ import cn.nukkit.timings.Timings;
 import cn.nukkit.utils.Binary;
 import cn.nukkit.utils.TextFormat;
 import cn.nukkit.utils.Zlib;
-
 import java.io.IOException;
 import java.nio.ByteOrder;
 import java.util.*;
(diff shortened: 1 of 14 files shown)


src/main/java/cn/nukkit/Player.java
```diff
@@ -3128,7 +3113,7 @@ public void handleDataPacket(DataPacket packet) {
                 case ProtocolInfo.REQUEST_CHUNK_RADIUS_PACKET:
                     RequestChunkRadiusPacket requestChunkRadiusPacket = (RequestChunkRadiusPacket) packet;
                     ChunkRadiusUpdatedPacket chunkRadiusUpdatePacket = new ChunkRadiusUpdatedPacket();
-                    this.chunkRadius = Math.max(5, Math.min(requestChunkRadiusPacket.radius, this.viewDistance));
+                    this.chunkRadius = Math.max(1, Math.min(requestChunkRadiusPacket.radius, this.viewDistance));
```
Pub4Game: It is not right, the minimum value is 5

src/main/java/cn/nukkit/utils/MainLogger.java
```diff
@@ -204,7 +204,7 @@ private String colorize(String string) {
     @Override
     public void run() {
         this.shutdown = false;
-        while (!this.shutdown) {
+        while (!this.shutdown && Nukkit.DEBUG > 0) {
```
Pub4Game: What? Noooo. It's not right. Better come up with some other solution.

minecrafter:
`HashMap<Long, T>` is very expensive in terms of memory usage. I suggest using GNU Trove instead.

Snake1999:
Nice! For some small problems mentioned by @pub4game, I'll check it out after my lunch (3 hours later in China) ;)

boy0001:
@jhelom Want to do a timings with and without this change?

PR #1076, opened 2016-10-04
Fix "invalid movement" when walking through doors

MrPowerGamerBR:
Movement fix from
https://github.com/ClearSkyTeam/ClearSky/commit/513b9dd38c0d4d1895d1bf9e4ce5add698017316


Snake1999:
Okay, this is a simple but important fix :P

PR #1091, opened 2016-10-05
Fix bow cannot be shot after sneaking/sprinting

MrPowerGamerBR:
From Genisys
https://github.com/iTXTech/Genisys/commit/dea6ecc73059bd9dd2ee4e59e5bfa16933250eb4

(Also, I always test my modifications (except those that I can't reproduce) before creating a pull request)

This has a issue because the stop/start sneak/sprint changes the player data flag to "No Action" so the server thinks we aren't pulling a bow anymore.

How to replicate:
1. Get a bow
2. Pull the bow but don't release
3. Sprint or sneak
4. Release the bow

The bow won't release a arrow, this fix changes the break to break the switch statement instead of breaking the current case.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -1741,6 +1741,7 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
                 return;
             }
 
+            packetswitch:
             switch (packet.pid()) {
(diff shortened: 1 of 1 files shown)


File14:
Great!

Snake1999:
:+1:

Muqsit:
Nice work!!

PR #1092, opened 2016-10-05
"Fix" ClassCastException when using EntityDamageByEntityEvent

MrPowerGamerBR:
This isn't a fancy fix, but I also have no idea why it thinks it should invoke EntityDamageEvent on a EntityDamageByEntityEvent. Probably later I will take a better look, but for now, this should work.


Snake1999:
Temporary fix here :)

PR #1093, opened 2016-10-05
Implement basic mob AI from pureentities

Creeperface01:

PR #1098, opened 2016-10-07
Revert "Implement basic mob AI from pureentities"

Snake1999:
Reverts Nukkit/Nukkit#1093
Our works should continue with more friendly API and multi-threading supports.
And for @Creeperface01 , we should write into EXISTING CLASSES, not creating lots of new classes. Java codes should written simple to be more effective.


Creeperface01:
Ok who'll implement it?

PR #1101, opened 2016-10-07
Typo fix

MrPowerGamerBR:


MrPowerGamerBR:
Yeah I should rename via Eclipse to fix classes that are still using the old name...

Issue #326, opened 2016-01-10
Regen chunks.


xpyctum:
![regen chunks](http://pp.vk.me/c631324/v631324517/ce10/WJme1O_gP_w.jpg)

Angelic47:
It also reproduced on my server. Maybe a bug with chunk reading.

caoli5288:
@Angelic47 Also maybe a bug with chunk save.

xpyctum:
@MagicDroidX @Angelic47 Please fix :(

MagicDroidX:
After all confirmed issues fixed, has this happened again?

xpyctum:
@MagicDroidX I will try, because this coming randomly

xpyctum:
@MagicDroidX still active....
![2016-02-11](https://cloud.githubusercontent.com/assets/6584811/12973541/e8b83a16-d0ce-11e5-909d-388f24139135.png)

xpyctum:
@MagicDroidX and now... still active. Now in mcregion. I think it's caused by autosaving, because it's happens only on those chunks, where equalizer
[https://pp.vk.me/c628026/v628026756/54f77/Nch6tOhjZhM.jpg](http://pp.vk.me/c628026/v628026756/54f77/Nch6tOhjZhM.jpg)

MagicDroidX:
please upload the level :)

xpyctum:
@MagicDroidX I think it helps.
[world.zip](https://github.com/Nukkit/Nukkit/files/179365/world.zip)

Pub4Game:
I think the problem is that the auto-save occurs when you change the chunk/unloading

xpyctum:
In my way, i can disable autosaving in-game process and enable it after game. I think it can fix error

Snake1999:
Does this issue happens again after https://github.com/Nukkit/Nukkit/pull/1404 ?

xpyctum:
@Snake1999 Need check this.

PeratX:
Any one can reproduce this using the latest Nukkit?

PR #1102, opened 2016-10-07
Multithreaded nukkit

boy0001:
This probably needs to be tested before it can be merged

Summary of changes:
- Refactor MainLogger to not break when used async (also faster)
- Switches to concurrent maps / atomic variables where useful
- Removes temporal locations as it isn't thread safe
- Adds temporal block method (no location, block id/data are safe)
- Optimize some existing code to use temporal block when block position isn't needed
- Optimize block cloning by removing some redundant instance fields in the block classes
- Removes level block cache as it's not thread safe
- Optimizes block get/set/relight significantly (even without the cache it's significantly faster)
- Removes level chunk map and use provider instead (obsolete having an extra map, and this reduces complexity)
- Optimizes chunk loading/saving (buffered reading, parallel save)
- Reduce some code duplication in the chunk classes
- Synchronizes chunk loading and saving to disk
- Changes to the chunk object are tracked by a LongAdder rather than a boolean
- Adds option to not save generated but unmodified chunks
- Synchronizes on inventory so it can accessed concurrently
- Refactors a lot of methods to work concurrently
- Reduces startup time by initializing the server in parallel
- Optimizes networking code (parallel compression)
- Optimizes packet encoding by initializing to the right size, and not array-copying on every written byte
- Optimizes login by removing artificial 200ms delay and correct chunk send order
- Improve scheduler by mapping the tasks rather than using an expensive priority queue
- Fix some block updates which don't need to be direct (instead send in bulk in parallel)
- Optimizes ChunkException by not calculating stacktrace (since the exception is used for logic flow)
- Tick the worlds in parallel
- Some changes to entities to work concurrently

Some stuff that may be included in the future:
- Config option to toggle parallel world ticking
- Move chunk loading/unloading and level gc to an entirely separate thread
- Fully porting existing code to use better methods (getTemporalBlock)
- Faster collision checking by not cloning the block instance


Pub4Game:
@Snake1999 Please don't add this, I need time to check

src/main/java/cn/nukkit/utils/TextFormat.java
```diff
@@ -109,7 +108,7 @@
 
     private static final Pattern CLEAN_PATTERN = Pattern.compile("(?i)" + String.valueOf(ESCAPE) + "[0-9A-FK-OR]");
     private final static Map<Integer, TextFormat> BY_ID = Maps.newTreeMap();
-    private final static Map<Character, TextFormat> BY_CHAR = new HashMap<>();
+    private final static Map<Character, TextFormat> BY_CHAR = new ConcurrentHashMap<>();
```
boy0001: It's thread safe and performs basically the same.

Creeperface01:
good 👍

File14:
oooooooooooo yeeeeeee ;D

src/main/java/cn/nukkit/utils/TextFormat.java
Snake1999: @pub4game hashmaps from java.util.current are thread safety. T hisminor change makes nukkit more compatible with multi threading :)

minecrafter: It would be better to make this immutable.

NycuRO:
@Snake1999 You can accept PR's? :)

Pub4Game:
Wait

MrPowerGamerBR:
I'm running my server with this commit and I didn't find any issues with it yet. :)

NycuRO:
@Snake1999 You can accept?

File14:
How many seconds reduce now

boy0001:
@File14 
Well nukkit already starts up for me in `0.8s`, now it takes `0.26s`. Though with a lot of plugins installed I suspect the it's not going to be that much of a difference.

boy0001:
Buffering the RandomAccessFile when reading chunks makes a huge difference, almost twice as fast now.

Creeperface01:
why don't you use constants? :D

boy0001:
With the `mightHaveMeta` check? I was lazy.

Creeperface01:
and what about getDrops() method can you change it to public Item[] getDrops() ? :D then we can drop item with compound tag :D

Creeperface01:
and i can't because your pull request is still opened :D

boy0001:
How would changing `int[][] getDrops()` to `int[] getDrops()` work?

Creeperface01:
no to Item[]

boy0001:
Oh, misread.

Creeperface01:
for example if chest block entity has custom name so if you break the block it should also drop item with custom name i think

iVertx:
screenshot_20161010-164156
screenshot_20161010-164231
screenshot_20161010-164255
I think here is some locks missing mainly in the RandomAcessFile implementation to prevent EOF while multiple threads are accessing the same region file.

MrPowerGamerBR:
@iVertx same thing happened to me in my world, I tried loading the world in Nukkit, OK, but after loading with this commit it detected so many corrupted chunks that it deleted some parts of my server spawn.

Creeperface01:
And will this be faster?

Creeperface01:
Because lots of servers have one main level and there is the most players

boy0001:
@Creeperface01 
Doesn't matter if you only have one level. Chunks are unloaded first, then it splits the level tick into 10 different tasks + each entity. The tasks are done in parallel. 

https://github.com/boy0001/Nukkit/blob/master/src/main/java/cn/nukkit/level/Level.java#L642

It'll perform better if the load is spread across the tasks (and you have a multi-core cpu). If there's some anomaly causing a lag spike in just a single task, it won't help with that.

To get the most out of it, the tasks which take the longest need to either be optimized or split up further. Some timings could help with this.

Creeperface01:
So this doesnt affect performance negatively?

Creeperface01:
Cant

Creeperface01:
is this still compatibile with master branch? and will it be merged?

minecrafter:
I've looked at this code. This is some of the most atrocious "multi-threaded" code I have ever seen. It looks as bad as some of the "async" projects I've seen to "speed up" Minecraft.

Making something properly multi-threaded, thread-safe and parallel does **not** mean liberally sprinkling `synchronized`, `volatile`, concurrent collections and `ForkJoinPools` everywhere. There are some particularly heinous cases: using concurrent collections when non-synchronized ones could have been used instead, and **even using a ForkJoinPool for array copies**.

@Creeperface01

Creeperface01:
i think so. I think we can make async login. it would be better

boy0001:
> This is some of the most atrocious "multi-threaded" code I have ever seen. It looks as bad as some of the "async" projects I've seen to "speed up" Minecraft.

Aside from my code style, the changes I've made result in measurable performance improvements. I can elaborate if you are going to deny this.
> liberally sprinkling synchronized

The cost of entering a synchronized block is pretty low in Java 8 unless there's contention. I don't see the problem of using it where concurrent access would cause issues?
> using concurrent collections when non-synchronized ones could have been used instead

Not a single case of this, unless you mean `ConcurrentHashMap` which isn't a collection, and doesn't perform noticeably slower than a HashMap.
> and ForkJoinPools everywhere

There's one ForkJoinPool, and it's used to tick the world in parallel. That's far from "everywhere".

> even using a ForkJoinPool for array copies.

No, you are mistaken. 

@minecrafter 
I am genuinely interested in some constructive feedback, if you do that sort of thing.

Creeperface01:
Idk but i think it wont be faster

minecrafter:
> Regardless of my code style, the changes I've made result in measurable performance improvements. I can elaborate if you are going to deny this.

Sure, the performance might be improved. But it's also unsafe code. Do you want a serious measurable performance improvement? Direct everything to `/dev/null`.

> The cost of entering a synchronized block is pretty low in Java 8 unless there's contention. I don't see the problem of using it where concurrent access would cause issues?

I never talked about the cost of using `synchronized`. I talked about the very liberal use of `volatile` and `synchronized`. Remember, what you're doing is "translating" non-thread-safe code to be vaguely thread-safe.

> Not a single case of this, unless you mean ConcurrentHashMap which isn't a collection, and doesn't perform noticeably slower than a HashMap.
`ConcurrentHashMap` implements `Map`, which is part of the Java Collections API. [The Javadoc says so.](https://docs.oracle.com/javase/7/docs/api/java/util/Map.html)

This is important: `ConcurrentHashMap` has elevated memory usage compared to `HashMap` (since maps are segmented). You almost never want `ConcurrentHashMap` unless you're looking for concurrent writes. Concurrent reads are always okay with `HashMap` since the map is never modified.

> No, you are mistaken.

8a786e61

Finally:

* [This](https://github.com/boy0001/Nukkit/blob/bb4ec51b8535775d6386ba130d327ede7a4f900f/src/main/java/cn/nukkit/plugin/PluginManager.java#L139) looks like a race condition.
* [This](https://github.com/boy0001/Nukkit/blob/bb4ec51b8535775d6386ba130d327ede7a4f900f/src/main/java/cn/nukkit/utils/MainLogger.java) is really bad. (To be fair, the old version was even worse.)

boy0001:
> Do you want a serious measurable performance improvement? Direct everything to /dev/null.

Pretty sure you'd get a lot of errors trying to read from a null directory. And it would probably worsen performance since reading from disk is a lot faster than using the CPU to generate a new overworld chunk.
> ConcurrentHashMap implements Map, which is part of the Java Collections API. The Javadoc says so.

So the Comparable class is also a collection since it's part of the Collections Framework? The collections FAQ clearly states that map isn't a collection:
http://docs.oracle.com/javase/8/docs/technotes/guides/collections/designfaq.html#a14
> This is important: ConcurrentHashMap has elevated memory usage compared to HashMap (since maps are segmented).

I'll change it, but it's mostly irrelevant since I've already set the segmentation to 1 in the constructor:
https://github.com/boy0001/Nukkit/blob/master/src/main/java/cn/nukkit/utils/TextFormat.java#L111
The actual overhead from segmenting the data (default=16) is pretty minuscule (probably less than 1%).
> 8a786e6

I've already changed that in this branch: https://github.com/boy0001/Nukkit/blob/master/src/main/java/cn/nukkit/utils/BinaryStream.java#L118
> This looks like a race condition.

How so? It fully loads each plugin (waits for the executor to finish) before loading the next series of plugins. Each time it only loads plugins which have all their dependencies loaded.
> This is really bad. (To be fair, the old version was even worse.)

The old version completely broke console when used async. I assume you're referring to my use of a lossy circular buffer which I used specifically so that it skips lines when it can't keep up. Disk usage can be pretty high for the logger, but I suppose that can be properly addressed in another PR.

Thanks for the feedback.

minecrafter:
> Pretty sure you'd get a lot of errors trying to read from a null directory. And it would probably worsen performance since reading from disk is a lot faster than using the CPU to generate a new overworld chunk.

I'm sure you didn't get my metaphor.

boy0001:
@minecrafter 
No, I get it, my code is unsafe, therefore equivalent to something nonfunctional. I just chose to interpret it literally.

minecrafter:
Unsafe != nonfunctional. Unsafe code could lead to unknown state, with potentially unlimited consequences.

Creeperface01:
Why you modified getBoundingbox() method? It was right

boy0001:
@Creeperface01 
It returns the same value, I just don't cache it since it's mostly unnecessary.

File14:
@Pub4Game @MagicDroidX Can you merge this now? I have tested this with SynapseAPI and it's fully working. I'm running it on my server now and there are no issues so far!

xpyctum:
> Optimizes chunk loading/saving (buffered reading, parallel save)
> Reduce some code duplication in the chunk classes
> Synchronizes chunk loading and saving to disk

Need check this :) 
I think it can fix this #326

boy0001:
@xpyctum 
Autosave occurs on the main thread, so it's probably something else.

Other things it could be:
 - You used the save command (console commands are async and therefore this will eventually cause chunk corruption)
 - You have a plugin modifying the world async.
 - There are bugs in how Nukkit saves or loads chunks which are unrelated to multi-threading

If this PR does help, it's probably the first or second thing.

xpyctum:
@boy0001 
>You have a plugin modifying the world async.

Yeah, my plugin uses async world edit.

NycuRO:
I have async world edit..

xpyctum:
@NycuRO I do not use your plugin, if that :)

NycuRO:
@xpyctum BiomeMaster not uses asyncs.

OverTook:
Is it deprecated?

yungtechboy1:
This is dead... Should Close?

OverTook:
@yungtechboy1 
agree

madecode15:
@boy0001 Is this dead?

Snake1999:
It's time to ramp down all PR and issues before our first release.
Fixing conflicts. @boy0001 , are you still here?

boy0001:
You can close this. I'll resubmit as multiple PRs at some point.

PR #1104, opened 2016-10-07
Add '-' as accepted username character

diamondo25:
Someone from iOS legitly tried to join with this character in her name. Lets accept it.
Android doesn't seem to accept it, but other characters are (like euro sign). Needs some investigation.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -1794,7 +1794,7 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
                         if ((c >= 'a' && c <= 'z') ||
                                 (c >= 'A' && c <= 'Z') ||
                                 (c >= '0' && c <= '9') ||
-                                c == '_' || c == ' '
+                                c == '_' || c == ' ' || c == '-'
                                 ) {
                             continue;
                         }


Muqsit:
Noooo.... This feature is right out of the blue. Even bukkit doesn't have it. Negative.

NycuRO:
Stupid ideea,please close :/

diamondo25:
> Noooo.... This feature is right out of the blue. Even bukkit doesn't have it. Negative.
This is more something for PE than the regular version of minecraft, so of course bukkit didn't have this (actually, the mojang account page restricts the use of characters in your player name)

src/main/java/cn/nukkit/Player.java
```diff
@@ -1794,7 +1794,7 @@ public void handleDataPacket(DataPacket packet) {
                         if ((c >= 'a' && c <= 'z') ||
                                 (c >= 'A' && c <= 'Z') ||
```
TheDiamondYT1: For the Nukkit devs (sorry for spam):

Why not use regex instead?

IrishUHC:
I support this change, - is an acceptable character in my opinion. Just because Bukkit doesn't have it, does it mean Nukkit can't?

TheDiamondYT1:
@IrishUHC I agreez

Snake1999:
As far as I know, many nukkit servers uses databases as well, and uses '-' to replace blanks in user names in convenience. If - is accepted to legal usernames, then we will have a big conflict problem. 

That's also why windows 10 not 9 after windows 8 - many softwares already detect number 9 for windows 95 98 and so on.

diamondo25:
Because other software that depends on us is written badly, we cannot do this? Thats a weird dependency. Can you give me an example plugi that does this?

Issue #1105, opened 2016-10-08
Item durability is broken


MrPowerGamerBR:
### Issue description
Okay, I don't know if this was already implemented in Nukkit, however I noticed this now and I thought I should report it.

### Steps to reproduce the issue
1. Get any tool
2. Try to use the tool
3. The durability bar doesn't decrease

### OS and versions

* Nukkit: Este servidor estß rodando Nukkit 1.0dev ???(Apple)?(Pie)? implementando versÒo da API 1.0.0 para Minecraft: PE v0.15.10 alpha (versÒo do protocolo 84)
* OS: Windows 10 and Android 6.0.1
* Device: PC and Moto G Turbo

### Crashdump, backtrace or other files
![http://i.imgur.com/l5gVsBW.gif](http://i.imgur.com/l5gVsBW.gif)
(look at the diamond shovel, also ignore that dirt, MCW10 issue)

MrPowerGamerBR:
Again, I don't remember if this was implemented before in Nukkit (even tho it should be very easy to implement if it isn't). No plugins, so it isn't a plugin fault.

Snake1999:
An unimplemented feature here :)

MrPowerGamerBR:
@Snake1999 seriously? I thought it was already implemented :yum:

Doesn't seem to difficult to implement tho, we already have the getMaxDurability() of every item (well, I hope getMaxDurability() does work, didn't try yet), just add 1 to the item durability on every tool use and check if the getMaxDurability() > durability, done.

Snake1999:
Try again as #1111 merged?)

kniffo80:
Can confirm it works. I think this one can be closed for now.

MrPowerGamerBR:
Can confirm, thanks @boy0001!

PR #1111, opened 2016-10-09
Fixes

### LINKED ISSUE #1108: Stuck at "Building Terrain" - Build 1375 or newer!
### Issue description
Commit that made it broke: https://github.com/Nukkit/Nukkit/commit/15e1b88e6e389ddf7453fa6ae6e90adc413c9b35
This is causing client hang on login on SOME old worlds. It gets stuck at "building terrain" but if you wait like 3 minutes you finally get into the server. See when it freezes: https://i.gyazo.com/dbfc5ada2f2c223650120495d50ef9b3.gif

You can download the following world and test by yourself. Make sure to go to X: 0 Y: 100 and Z: 0
for the chunk to load. If you spawn in that position it will freeze at "building terrain" otherwise if you spawned in another place and go to the locaiton it will freeze your client. 

Download world: https://www.dropbox.com/s/6z761f16rf0vx35/world.zip?dl=1

### OS and versions

* Nukkit: Build 1375 - https://circleci.com/gh/Nukkit/Nukkit/1375
* OS: Host in on Linux but doesn't matter in this case.
* Device: Windows 10 but doesn't matter in this case.

### Crashdump, backtrace or other files
There are no sign of errors or bug reports.

**INFO**
This is @boy0001's commit so I suggest him to take a look at it.
@MrPowerGamerBR tested my world and is also having the same issuse with @boy0001's commit!

boy0001:
Fixes #1108
Fixes plugin loader not recovering from class not found error
Fixes tool durability

### DIFF
--- a/src/main/java/cn/nukkit/item/ItemTool.java
+++ b/src/main/java/cn/nukkit/item/ItemTool.java
@@ -2,13 +2,9 @@ package cn.nukkit.item;
 
 import cn.nukkit.block.Block;
 import cn.nukkit.entity.Entity;
-import cn.nukkit.event.entity.EntityDamageEvent;
 import cn.nukkit.item.enchantment.Enchantment;
-import cn.nukkit.item.enchantment.EnchantmentDurability;
-import cn.nukkit.item.enchantment.EnchantmentFireAspect;
 import cn.nukkit.nbt.tag.ByteTag;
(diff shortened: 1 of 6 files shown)


Snake1999:
I'll add a note that #1105 is also fixed here
Awesome ;)

File14:
YES, Thanks for fixing it!

PR #1112, opened 2016-10-09
minor binary stream optimization

boy0001:
- Add a new buffer instead of resizing every write
- Join the buffers in parallel when requested
- Use buffer for common read sizes (short, int, long)


Snake1999:
Well done :)

PR #1116, opened 2016-10-09
Finally fix Memory leak and Light Cave and Lights in general [BUG]

NycuRO:
Commited by thebigsmile ,from ClearSky.

### DIFF
--- a/src/main/java/cn/nukkit/block/Block.java
+++ b/src/main/java/cn/nukkit/block/Block.java
@@ -511,7 +511,7 @@ public abstract class Block extends Position implements Metadatable, Cloneable {
                     transparent[id] = block.isTransparent();
                     hardness[id] = block.getHardness();
                     light[id] = block.getLightLevel();
-                    lightFilter[$id] = 1;
+                    lightFilter[id] = 1;
 
                     if (block.isSolid()) {
                         if (block.isTransparent()) {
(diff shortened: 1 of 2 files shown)


Snake1999:
Oh, it says `All checks have failed`. Could you try compile again?)

NycuRO:
@Snake1999 Ready :)

NycuRO:
@Snake1999 Accept?

Pub4Game:
It's so brilliant what is NOT.

You do realize that it changes nothing?

NycuRO:
@Pub4Game test.

Pub4Game:
🤔

NycuRO:
@Snake1999 You can accept? Is correct.

MagicDroidX:
I cannot see the real point. There is no possibility of memory leak according to the original codes and the modified one.

NycuRO:
@MagicDroidX Code is from ClearSky. Fixed this issue from Genisys ,and Nukkit have: https://github.com/iTXTech/Genisys/issues/1788

MagicDroidX:
There is no reference to show that these codes cause memory and as a said there is NO possibility to have a memory leak according to these codes. From my point of view, these changes have no efforts but just a possibility to break the logic.

NycuRO:
@MagicDroidX Yes ,but this bug fix and Light in cave and light in general..

MagicDroidX:
Well, in that case, the title is really misleading and you should edit.
And I will look into it. :)

NycuRO:
@MagicDroidX Ready.

MagicDroidX:
To be frank, there is no difference at all. :/

NycuRO:
@MagicDroidX It's.

NycuRO:
Delete world,create new and watch caves

Snake1999:
The logic in the code is not changed, that means, this PR changed nothing. The code is not better for developers to read, that's what we aren't looking forward to.

PR #1119, opened 2016-10-09
player.setGamemode(int gamemode, AdventureSettings newSettings);

boybook:


Pub4Game:
What!?

boybook:
I kept the old API.
player.setGamemode(int gamemode);

Snake1999:
This PR fixes a very tiny but important bug.
When your internet are not good enough (e.g. excceeding the speed limit), UDP packets sometimes are not sent in correct order. If some player is set to adventure mode with flights enabled, and he should be set to survival without flights, it should be:
```
set gamemode -> set allow flights -> set gamemode
```
but with sheepy internet, packets are sometimes in incorrect order, what client received might be:
```
set gamemode -> set gamemode -> set allow flights
```
By this, he/she is set to survival gamemode, but he/she could fly anyway. That's where the problem is.

With this fix, AdventureSettings and gamemodes are bundled together, which preventing the small problem to happen. It seems little to survival servers, but is meaningful to minigame servers.

MrPowerGamerBR:
@Snake1999 thanks for explaning 👍

PR #1120, opened 2016-10-09
Crafting fix (don't check inv if it contains air)

boy0001:


Pub4Game:
Tested?

boy0001:
Yes, seems to work.

NycuRO:
@Snake1999 Accept? Tested and work.

PR #1122, opened 2016-10-09
Binary stream optimizations

boy0001:


Pub4Game:
Please test

boy0001:
I had tested this with synapse

File14:
And it worked to login to the server without getting stuck at "building terrian" and then get timeout?

src/main/java/cn/nukkit/utils/BinaryStream.java
```diff
@@ -59,113 +74,221 @@ public void setOffset(int offset) {
...
+            this.buffer = data;
+            this.buffers.clear();
+            return this.buffer;
+        }
+        final byte[] data = new byte[getCount()];
+        // Check if we have a list of buffers
+        int pos = 0;
+        for (final byte[] bytes : buffers) {
+            final int finalPos = pos;
+            POOL.submit(new Callable() {
```
minecrafter: Does an array copy _really_ need a `ForkJoinPool`? I'd imagine this would be slower than just doing it sequentially.

NycuRO:
@Snake1999 Tested. You can accept?

PR #1124, opened 2016-10-09
Add PlayerInteractEvent(...) when a player is trampling a crop

MrPowerGamerBR:


File14:
Great! YES

Pub4Game:
No. it should be in PlayerInteractEvent - Physical

MrPowerGamerBR:
@Pub4Game can I create a new pull request changing to PlayerInteractEvent?

Also, can't mobs also trample crops?

Pub4Game:
You can re-make it in this pull request

MrPowerGamerBR:
@Pub4Game so... I will need to check if it is a Player to then create the PlayerInteractEvent, right? Because entities can also trample crops by jumping.

Also, I think crop trampling is a bit broken, the event is only triggered if you jump from a height higher than 1 block, anything less than that doesn't trigger. (Which, IIRC, isn't the correct behavior)

File14:
@Pub4Game Why not having it this way? :)

MrPowerGamerBR:
@Pub4Game done

File14:
How to cancel now? @MrPowerGamerBR

MrPowerGamerBR:
@File14 

Same way as you would do in Bukkit

File14:
But via Block.FARMLAND?

MrPowerGamerBR:
@File14 yep

Issue #420, opened 2016-01-27
Change how the "invalid movement" works.


MrPowerGamerBR:
The invalid movement check, while useful, affects a lot of simple vanilla functions.

**Example:**
- Sprinting on a slab causes rubberbanding
- Recieving damage causes rubberbanding

Yes, I know that you can disable the check, but it would be better to just fix the check than disable it. :+1:

Pub4Game:
:+1:

Snake1999:
:+1:

xxFlare:
@MrPowerGamerBR Yes, Invalid movement does not take sprinting into consideration because it was created in MCPE 0.10 (based off PocketMine).

Though disabling it will allow a lot of hacks on the server such as NoClip.

MrPowerGamerBR:
Yeah, disabling invalid movement is totally a bad option, but the invalid
movement check must be changed to not affect simple things like sporting or
damage knockback.
Em 8 de fev de 2016 11:34, "xFlare" <notifications@github.com> escreveu:

Guillaume351:
How to disable it ? I allowed flight in the config, but still rubberband ;/ Thanks

Pub4Game:
@MagicDroidX Please fix it. This is a very important issue.

Pub4Game:
I think this is what you need.
```
@EventHandler(priority = EventPriority.NORMAL, ignoreCancelled = false)
public void onPlayerInvalidMoveEvent(PlayerInvalidMoveEvent event) {
        event.setRevert(false);
}
```

Tee7even:
Nope, he doesn't request to disable it, but to fix how it works.

Pub4Game:
Oops.. I wanted to reply to the last comment

PR #1128, opened 2016-10-09
Make invalid movement check more lenient when walking up stairs/slabs

MrPowerGamerBR:
Instead of applying the default 0.125 value on the movement check, the
code first checks if there is a stair or slab nearby, if yes, the value
is increased to 0.19 (if he isn't sprinting) or 0.32 (if he is
sprinting).

I used 0.19 and 0.32 because these were the max value I got in a vanilla
MCW10 client when trying to walk on a stair.

After walking in a stair up and down for hours, I think the check is now
working better than before, with less rubberbanding issues.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -3,6 +3,8 @@ package cn.nukkit;
 import cn.nukkit.block.Block;
 import cn.nukkit.block.BlockAir;
 import cn.nukkit.block.BlockDoor;
+import cn.nukkit.block.BlockSlab;
+import cn.nukkit.block.BlockStairs;
 import cn.nukkit.blockentity.BlockEntity;
 import cn.nukkit.blockentity.BlockEntityItemFrame;
 import cn.nukkit.blockentity.BlockEntitySign;
@@ -1194,6 +1196,34 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
         }
     }
 
+    protected boolean shouldMakeCheckMoreLenient(Block b) {
+        Block check = null;
+        if (b instanceof BlockStairs || b instanceof BlockSlab) {
+            return true;
+        }
+        check = b.getSide(Vector3.SIDE_DOWN);
+        if (check instanceof BlockStairs || check instanceof BlockSlab) {
+            return true;
+        }
+        check = b.getSide(Vector3.SIDE_EAST);
+        if (check instanceof BlockStairs || check instanceof BlockSlab) {
+            return true;
+        }
+        check = b.getSide(Vector3.SIDE_NORTH);
+        if (check instanceof BlockStairs || check instanceof BlockSlab) {
+            return true;
+        }
+        check = b.getSide(Vector3.SIDE_SOUTH);
+        if (check instanceof BlockStairs || check instanceof BlockSlab) {
+            return true;
+        }
+        check = b.getSide(Vector3.SIDE_WEST);
+        if (check instanceof BlockStairs || check instanceof BlockSlab) {
+            return true;
+        }
+        return false;
+    }
+    
     protected void processMovement(int tickDiff) {
         if (!this.isAlive() || !this.spawned || this.newPosition == null || this.teleportPosition != null) {
             return;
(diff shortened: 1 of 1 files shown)


MrPowerGamerBR:
Fixes #420

File14:
Thanks!

Pub4Game:
It's awful

MrPowerGamerBR:
@Pub4Game at least it is not rubberbanding like crazy when walking up stairs.

I also don't know how to improve the check, if you have a better idea, please improve it :)

MrPowerGamerBR:
@Pub4Game also, how do you think the check could be improved?

The problem is that rubberbanding happens, that's the problem, this happens when trying to walk stairs or walk in slabs, and the problem is even more common if you are walking and then you quickly turn our vision around.

The check is bad because it is checking a region 2x2x2 (?) around the player

```
 X 
XPX
 X
```

The X are the checked blocks and the P is the player, and, of course, the block below the player and the block at the player is also checked.

The check also doesn't account for sprinting, which also makes the rubberbanding issue more visible, on a vanilla Minecraft Windows 10 Client, I got those two values analyzing the ```diff``` from trying to climb stairs.

Walking: 0.19

Spriting: 0.32

So the only way to reduce the rubberbanding is by increasing the diff check value, however increasing the diff value to 0.32 won't be as good, since you are allowing more invalid movement hacks, so the only way is to check the block that the player is above to then change the diff check value, however there are two problems: 

1. The block sometimes is at the player, not below it. (If you do getBlock() while in a stair, it will return the stair)
2. The block sometimes is in front of the player. (If you do getBlock() while in a stair, it will return air)

The first one is easy, just check if the block at the player is a stair, the second one... yeah, it isn't easy because the player's yaw need to be checked (by checking if the yaw is between 0 and 90... check the block on north, etc etc etc... checking diagonals needs to be done too), so, while I don't do this, I done that not-so-good check, that works but... it isn't perfect.

That's why I done this pull request, this issue is in Nukkit since January, when I first reported this issue, 9 months has past and the issue still persists, so I created my own solution while a better fix isn't implemented in Nukkit.

Creeperface01:
Nooo

Creeperface01:
It slow down a server

Creeperface01:
And i think we can find another solution

MrPowerGamerBR:
Then the only solution is by bumping up the diff value check to something
waaaay to high, which isn't the best solution.

Or rewriting the entire check to work properly.

Em 10 de out de 2016 2:53 AM, "Creeperface01" <notifications@github.com>
escreveu:

MrPowerGamerBR:
But if it is causing lag issues, then it should be reverted ASAP.

Em 10 de out de 2016 6:37 AM, "Leonardo Malaman" <leonardomalaman@gmail.com>
escreveu:

MagicDroidX:
How about this:
Only if the difference is higher than it is in normal cases shall we try to find steirs or slab around the player.

MrPowerGamerBR:
Yeah, that's a better idea ;)

Another idea I had was only blocking movement if the player moved wrongly
twice in less than 3 seconds.

Em 10 de out de 2016 8:26 AM, "MagicDroidX" <notifications@github.com>
escreveu:

Creeperface01:
Or why we can not save blocks around player and use them in collision checks or onGround check?

Creeperface01:
i'll make pull request with better collision checking

Creeperface01:
#1134 then you can get these blocks faster

MagicDroidX:
Conclusion:
1. Only get blocks when `diff` is bigger than 1.25
2. Get blocks around player by using `getBlockId`. It doesn't affect the efficiency.
3. Maybe we can add a counter to revert player if he is judged to move wrongly.

MrPowerGamerBR:
@MagicDroidX yeah that would be better, also I didn't know that using getBlockIdAt(...) is better than using getBlock(...)

boy0001:
@MrPowerGamerBR 
getBlock(...) has to clone a new Block object each time, set the location for it, check/add to the cache, get the block data and nbt (if applicable). It's much slower.

File14:
Any process? @MrPowerGamerBR

Creeperface01:
what exactly does it check?

PR #1134, opened 2016-10-10
Better collision checking

Creeperface01:
i named this method getCollisionBoundingBox() because idk how should i name it :D so you can rename it or tell me better name


Creeperface01:
wait

MagicDroidX:
`getBlock` is a time-consuming method so in my opinion, this is undoubtedly slower.

Creeperface01:
its slower if move event is canceled but its not very often

Creeperface01:
or what should we use instead of getBlock()? getBLockIdAt() ?

Creeperface01:
it won't be slower than now

MagicDroidX:
I've made a conclusion here: https://github.com/Nukkit/Nukkit/pull/1128
PTAL

Creeperface01:
OK but with this change you can check blocks without affecting performance too. Because blocksAround or blockUnder lists contains only few blocks

Creeperface01:
but in entity base tick we even get these blocks so i dont know why should be this slow

Creeperface01:
i removed underBlocks, but if someone would want to get them in playerMOveEvent, he can use getBlocksUnder() and in base tick nukkit won't find these blocks again

src/main/java/cn/nukkit/Player.java
```diff
@@ -1849,6 +1851,7 @@ public void handleDataPacket(DataPacket packet) {
 
                         this.setRotation(movePlayerPacket.yaw, movePlayerPacket.pitch);
                         this.newPosition = newPos;
+                        this.newPosition = newPos;
```
Creeperface01: WTF :D thx

Creeperface01:
Why?

PR #1136, opened 2016-10-10
Fixed brewing

Creeperface01:


NycuRO:
Tested. Work. You can accept? @Snake1999

Pub4Game:
@NycuRO If flood will continue you'll be blocked in this repository.

NycuRO:
Ok sorry ,but is for more speed PR answers. Not all PR isn't accepted because authors not have one "clock" for avertise..

MagicDroidX:
Thanks for your work on brewing.
Have you tested this? :)

Creeperface01:
Sure, works

MagicDroidX:
Any side effect?

Creeperface01:
Probably not, but i think it would be faster if we store cook time etc in variables instead of compoundtag

MagicDroidX:
Okay i see.

PR #1137, opened 2016-10-10
More in Anvil

NycuRO:


NycuRO:
@Creeperface01 It's correctly?

src/main/java/cn/nukkit/inventory/AnvilInventory.java
```diff
@@ -44,9 +44,9 @@ public boolean onRename(Player player, Item resultItem) {
         }
 
         if (local.getId() != 0 && second.getId() == 0) { //only rename
+            player.getInventory().addItem(local);
```
NycuRO: @Creeperface01 #Blame_Creeper. First is this for fix bug of duplicate item ;/

src/main/java/cn/nukkit/item/Item.java
```diff
@@ -1342,6 +1342,33 @@ public static Item get(int id, Integer meta, int count, byte[] tags) {
...
+    }
+
+    public static Item send(int id, Integer meta) {
+        return send(id, meta, 1);
+    }
+
+    public static Item send(int id, Integer meta, int count) {
+        return send(id, meta, count, new byte[0]);
+    }
+
```
Pub4Game: what? it very impractical

Pub4Game:
You broke the renaming of the item because it is issued before setCustomName()

Creeperface01:
I or somebody else will improve anvil soon

PR #1103, opened 2016-10-07
Fixes a DoS with empty packets

minecrafter:
Nukkit does not properly process empty packets. Since only a few packets at a time are processed, sending a sufficiently large number of them can freeze the RakLib thread and cause a denial of service.

See also pmmp/RakLib@bab1c10f1556009574ed7adc4deeaede284f1b2d

PR #1138, opened 2016-10-11
Fix query

MrPowerGamerBR:
I should be studying for ~~school~~ high school, but no... I'm here sending pull requests to Nukkit because Minecraft is *so* better than studying... *just kidding*

Anyway, readding https://github.com/Nukkit/Nukkit/pull/1103/files#diff-a4147b106d2a957465633d3a045c9db1L187 fixes query, thanks @minecrafter for showing what should be readded 👍 

```
{
    "status": true,
    "version": "v0.15.10 alpha",
    "players":
     {
        "online": 0,
        "max": 20
    }
}
```


MagicDroidX:
:+1:

PR #1141, opened 2016-10-12
Implements more for generator

NycuRO:
Implements BlockPos, Vec3i and other. Is needed for generator extends :)


NycuRO:
Create new.

PR #1132, opened 2016-10-09
Add BlockFormEvent when a obsidian or cobblestone is generated from liquid

MrPowerGamerBR:
Something @imjack wanted and I thought it was a good idea to implement to Nukkit's API.

**Note:** Bukkit doesn't use BlockFormEvent for Obsidian/Cobblestone generated from liquids, however Nukkit doesn't have a BlockFromToEvent, so I used BlockFormEvent, if using BlockFromToEvent is better, then ask that I will change the code.

PR #1143, opened 2016-10-12
Implements BlockFromToEvent

imjack:
As seen in pull request #1132 . Not to sure if this is correct but it does what its suppose to..

### DIFF
--- a/src/main/java/cn/nukkit/block/BlockLiquid.java
+++ b/src/main/java/cn/nukkit/block/BlockLiquid.java
@@ -1,15 +1,17 @@
 package cn.nukkit.block;
 
+import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+
+import cn.nukkit.Server;
 import cn.nukkit.entity.Entity;
+import cn.nukkit.event.block.BlockFromToEvent;
 import cn.nukkit.item.Item;
 import cn.nukkit.level.Level;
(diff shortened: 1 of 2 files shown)


imjack:
I'm not too sure why the tests failed.. It compiles fine on my PC without errors.

src/main/java/cn/nukkit/block/BlockLiquid.java
```diff
@@ -193,6 +195,8 @@ public int tickRate() {
 
     @Override
     public int onUpdate(int type) {
```
NycuRO: public int onUpdate(int type, BlockFromToEvent event) { -> for fix

src/main/java/cn/nukkit/block/BlockLiquid.java
```diff
@@ -247,10 +251,25 @@ public int onUpdate(int type) {
 
                 if (k != decay) {
                     decay = k;
+                    Block b;
```
NycuRO: if (k != decay, Block b) {

Snake1999:
Your commit won't be merged as:

Git ignore rule edited
Language submodule deleted
One or more checks failed

imjack:
I'll look into it

src/main/java/cn/nukkit/block/BlockLiquid.java
TheDiamondYT1: No

PR #1149, opened 2016-10-15
Basic entity pathfinding

boy0001:
Hostile mobs will traverse around obstacles to attack the player (only hostile mob is creeper right now)

Example usage:

``` Java
MobAI a = new NeutralMobAI(this, 16, 16, 0);
MobAI b = new FollowOwnerAI(this, 128);
MobAI c = new FollowItemAI(this, Item.BONE, 32, 16);
MobAI d = new CreatureWanderAI(this);
entity.setAI(new MobAIUnion(a, b, c, d));
```

### DIFF
--- a/src/main/java/cn/nukkit/entity/ai/MobAIUnion.java
+++ b/src/main/java/cn/nukkit/entity/ai/MobAIUnion.java
@@ -1,0 +1,55 @@
+package cn.nukkit.entity.ai;
+
+import cn.nukkit.entity.Entity;
+import cn.nukkit.math.Vector3;
+
+/**
+ * You can combine different AI's together
+ *  - The fist MobAI to return a valid target is used
+ */
+public class MobAIUnion implements MobAI {
+
+    private final MobAI[] tactics;
+    private MobAI currentTactic;
+
+    public MobAIUnion(MobAI... ai) {
+        this.tactics = ai;
+    }
+
+    @Override
+    public boolean isValidTarget(Vector3 target) {
+        for (MobAI tactic : tactics) {
+            if (tactic.isValidTarget(target)) {
+                currentTactic = tactic;
+                return true;
+            }
+        }
+        return false;
+    }
+
+    @Override
+    public Vector3 findTarget() {
+        for (MobAI tactic : tactics) {
+            Vector3 target = tactic.findTarget();
+            if (isValidTarget(target)) {
+                return target;
+            }
+        }
+        return null;
+    }
+
+    @Override
+    public Vector3 calculateNextMove(Vector3 target) {
+        if (currentTactic != null) {
+            return currentTactic.calculateNextMove(target);
+        }
+        return null;
+    }
+
+    @Override
+    public void processAttacker(Entity attacker) {
+        for (MobAI tactic : tactics) {
+            tactic.processAttacker(attacker);
+        }
+    }
+}
--- a/src/main/java/cn/nukkit/block/Block.java
+++ b/src/main/java/cn/nukkit/block/Block.java
@@ -13,7 +13,6 @@ import cn.nukkit.metadata.MetadataValue;
 import cn.nukkit.metadata.Metadatable;
 import cn.nukkit.plugin.Plugin;
 import cn.nukkit.utils.BlockColor;
-
 import java.lang.reflect.Constructor;
 import java.util.List;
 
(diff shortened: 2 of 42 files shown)


yescallop:
Bugs found:
 - Wolf causes client crashing
 - Entity sometimes stuck in a corner

Details:
 - Entity cannot float on water

boy0001:
@yescallop 
Jump height was too low - fixed
Seems the client doesn't recognize the nbt related to tamed entities, so I commented that out

yescallop:
Jump height too high that jumping onto a fence is possible

yescallop:
Knocking back an entity looks strange

yescallop:
Falling is no damage

boy0001:
> Entity cannot float on water
Fixed

> Jump height too high that jumping onto a fence is possible
The pathfinder should determine where it jumps, so jump height shouldn't matter.

> Knocking back an entity looks strange
It's always looked strange

> Falling is no damage
Yes there is?

yescallop:
https://github.com/Nukkit/Nukkit/pull/1149/commits/ab56577f2ff51adb3ef13dd17489319e18c28df8
Seems this causes a crash

boy0001:
> Seems this commit causes a client crash

That commit just changes the motionX/Y/Z values for the Entity, it wouldn't make sense for it to cause a crash, and it's not crashing at all for me?

If you have any wolves from earlier you need to delete them as Nukkit's EntityTameable class caused the crash there and I had to comment it out.

yescallop:
But you've fixed it, confirmed the crash isn't caused by https://github.com/Nukkit/Nukkit/pull/1149/commits/ab56577f2ff51adb3ef13dd17489319e18c28df8

kniffo80:
The tamed stuff was copied from original PC server version. Just checked that and tried a lot of stuff with taming. It seems, that the data structure for MCPE is another one than on PC version. I took a look on a lot of other implementations - but none found out how taming is implemented.

A decompiled android client would be a good idea (but i think only the library which is - i assume at this point - written in C++). Then we can take a look how it's implemented.

src/main/java/cn/nukkit/entity/ai/MobAIUnion.java
```diff
@@ -0,0 +1,55 @@
+package cn.nukkit.entity.ai;
+
+import cn.nukkit.entity.Entity;
+import cn.nukkit.math.Vector3;
+
+/**
+ * You can combine different AI's together
+ *  - The fist MobAI to return a valid target is used
+ */
+public class MobAIUnion implements MobAI {
```
liach: Well I assume that you can create some class like `MobAISet` for the AI for a mob.

Creeperface01:
@boy0001 awesome! but why did you add getFloorX() to block class? its already in Vector3 class :D

IrishPacks:
I seem to be getting errors.

22:48:05 [ALERT] java.lang.NullPointerException
        at java.io.Reader.<init>(Unknown Source)
        at java.io.InputStreamReader.<init>(Unknown Source)
        at cn.nukkit.utils.Utils.readFile(Utils.java:72)
        at cn.nukkit.lang.BaseLang.loadLang(BaseLang.java:88)
        at cn.nukkit.lang.BaseLang.<init>(BaseLang.java:37)
        at cn.nukkit.lang.BaseLang.<init>(BaseLang.java:29)
        at cn.nukkit.lang.BaseLang.<init>(BaseLang.java:25)
        at cn.nukkit.Server.<init>(Server.java:278)
        at cn.nukkit.Nukkit.main(Nukkit.java:65)

IrishPacks:
One problem I find with this is that the mobs move too fast and don't look the direction they're going and oh boy do they move far when you hit them. but this is just the basics so good job :)

boy0001:
@Creeperface01 
The getFloorX/Z was broken for entities (returned wrong value for negative coords). I fixed it, but figured it would be faster to just cast to int for blocks since they always have an integer value.

@IrishPacks 
Error isn't related to this at all. In the nukkit.yml reset the `lang` to `eng` or something, or just delete the whole file and it should fix it.

And yeah, I haven't touched the default speed for entities (not really sure what the right values would have been).

IrishPacks:
Yeah i fixed the lang issues, I'll mess around with the speed values and get back to you :)

Creeperface01:
Use values from pureentities

Creeperface01:
so when this will be ready to merge? :D

File14:
When will this be ready? @boy0001

Creeperface01:
@boy0001 can you update it? :D

Creeperface01:
@boy0001 or if you could create patch it would be good

Creeperface01:
im testing it and looks good, only mobs speed is too high :D and move duration too short

Creeperface01:
so its only about values now i think

Creeperface01:
maybe entity.getSpeed() / 20 because speed is in blocks per second

File14:
It would be cool if we could add this! Finally moveable mobs :D

PeratX:
Outdated.

PR #1150, opened 2016-10-16
fix color on desert biome

NycuRO:


NycuRO:
Not tested. Need test. If not work,close.

Creeperface01:
but color is calculated from temperature and railFall isn't it?

Pub4Game:
@Creeperface01 You're right

Creeperface01:
and desert biome temperature and rainfall are right

Creeperface01:
@Pub4Game and why you don't accept/decline new pull requests? :D then we can not create new, because they won't be compatibile :D

NycuRO:
Yes.. accept please or close. @Pub4Game And sorry for flood you.

PR #1151, opened 2016-10-16
kick joining player with the existing name

Creeperface01:
i think we should kick joining player instead of already playing one


Pub4Game:
No, you break plugins API

Creeperface01:
why?

Creeperface01:
and plugins should check if message.equals("Logged in from another location")

MrPowerGamerBR:
This isn't the default behavior in MCPE and isn't the default behavior in Minecraft PC, if another player logs in into the server and another is already logged in, the current logged in player is kicked.

Pub4Game:
Ohh... Please read https://github.com/Nukkit/Nukkit/pull/893/files

PR #1153, opened 2016-10-17
Fix gamemode command as well as add an alias

IrishPacks:
Fix some things inside of /gamemode as well as add an alias for convenience.

### DIFF
--- a/src/main/java/cn/nukkit/command/defaults/GamemodeCommand.java
+++ b/src/main/java/cn/nukkit/command/defaults/GamemodeCommand.java
@@ -16,6 +16,7 @@ public class GamemodeCommand extends VanillaCommand {
     public GamemodeCommand(String name) {
         super(name, "%nukkit.command.gamemode.description", "%commands.gamemode.usage");
         this.setPermission("nukkit.command.gamemode");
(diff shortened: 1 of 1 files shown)


Pub4Game:
Don't see the point in the changes that you made. Game mode in any case be changed, even though the message suggests otherwise.

IrishPacks:
Before it would always display the message that their gamemode was changed even if they were in the same gamemode, thats a fix. test yourself. do /gamemode c and then do it again, it'll always keep switching your gamemode. I had to reposition where the gamemode was switched as it was affecting getting their proper gamemode and it should be put last anyways.

Creeperface01:
it doesnt, it checks if you already haven't the same gamemode in setGamemode() method

IrishPacks:
Try it for yourself. it keeps saying the gamemode has updated even if you are in the same gamemode.

boy0001:
@Creeperface01 @MagicDroidX 
The result of setGamemode is ignored entirely in the gamemode command. 
https://github.com/nukkit/Nukkit/blob/master/src/main/java/cn/nukkit/command/defaults/GamemodeCommand.java#L56

It does send the wrong message.

PR #1162, opened 2016-10-19
Debug stuff

boy0001:
Adds debug command which pastes some useful information for bug reports
Debug commands should be enabled by default
Example: http://hastebin.com/witihopira

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,18 +1,18 @@
 ### Issue description
-<!--- use our forum https://forums.nukkit.io for questions -->
+<!--- Feature requests can ignore this template -->
+<!--- Use our forum https://forums.nukkit.io for questions -->
 Write a short description about the issue
 
 ### Steps to reproduce the issue
-<!--- help us find the problem by adding steps to reproduce the issue -->
+<!--- Help us find the problem by adding steps to reproduce the issue -->
 1. ...
 2. ...
 
-### OS and versions
-<!--- use the 'version' command in Nukkit -->
-* Nukkit:
-* OS:
-* Device:
+### Debug information
+<!--- Use the 'debug' and 'timings paste' command in Nukkit -->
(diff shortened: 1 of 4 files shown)


PeratX:
Debug commands shouldn't be default enabled.

boy0001:
@PeratX 
But why? Disabling does nothing except makes Nukkit harder to use? Please give me one good reason why `/status` needs to be disabled.

Snake1999:
@boy0001 we are developers who will think debug commands should be enabled in normal. But all of our users come from everywhere on the Earth, speak more than 11 languages as we support and translated. Debug commands are only in English, and only developers can get useful information from it.
You can add it back, or let we add it after merged.

boy0001:
@Snake1999 
Why should we care whether they might understand it or not? This PR specifically adds a debug command which automatically uploads all the information potentially relevant for someone to create a ticket. It makes it a lot easier for people to report issues, and a lot easier to resolve them, but you want to have the commands disabled.

I don't get it? If they never have to use it (or understand), good, but disabling it just makes things more difficult.

PeratX:
@boy0001 @Snake1999 is right. Not all users are developers, they don't need these commands.

TheDiamondYT1:
Who cares if they want it? Its a useful command and it doesnt have to be touched unless they are creating an issue

MrPowerGamerBR:
I don't know why "/status" needs to be enabled via a config switch, I understand why blocking "/gc" is good, but "/status"? Status is harmless and, in my opinion, very useful for a server owner.

TheDiamondYT1:
Approve the changes

Pub4Game:
No. It will be rejected

TheDiamondYT1:
@Pub4Game Why

TheDiamondYT1:
Well dont complain when people dont give enough info on issues

boy0001:
@Snake1999 Can this be merged if I make the command disabled, or possibly some other changes?

PR #1165, opened 2016-10-19
Default value should be true

Creeperface01:
in method name is "add" so i suppose, default value should be true


MagicDroidX:
More description on this please.
I don't get the point.

Creeperface01:
Its like event.setCancelled() there is optional boolean variable but default is true.. Its the same here

MagicDroidX:
What's the purpose to replace `null` with `true`?

Creeperface01:
Null = false

Creeperface01:
@MagicDroidX so?

PR #1167, opened 2016-10-20
Change kick API

boy0001:
Keeps old methods for compatiblity, but all existing code now uses the
PlayerKickEvent.Reason so it can be properly identified in an event.

### DIFF
--- a/src/main/java/cn/nukkit/Server.java
+++ b/src/main/java/cn/nukkit/Server.java
@@ -16,6 +16,7 @@ import cn.nukkit.entity.weather.EntityLightning;
 import cn.nukkit.event.HandlerList;
 import cn.nukkit.event.level.LevelInitEvent;
 import cn.nukkit.event.level.LevelLoadEvent;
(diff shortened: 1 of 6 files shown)


PeratX:
seems nice!

src/main/java/cn/nukkit/Server.java
```diff
@@ -908,9 +909,10 @@ public void sendRecipeList(Player player) {
 
     private void checkTickUpdates(int currentTick, long tickTime) {
         for (Player p : new ArrayList<>(this.players.values())) {
-            if (!p.loggedIn && (tickTime - p.creationTime) >= 10000) {
-                p.close("", "Login timeout");
-            } else if (this.alwaysTickPlayers) {
+            if (!p.loggedIn && (tickTime - p.creationTime) >= 10000 && p.kick(PlayerKickEvent.Reason.LOGIN_TIMOUT)) {
+                continue;
```
Pub4Game: close()

PR #1178, opened 2016-10-22
Revert "Binary stream optimizations"

File14:
Reverts Nukkit/Nukkit#1122

Because this breaks Synapse when logging into a server. I crash when the chunks are being loaded sort of.


NycuRO:
I don't have this issue.

File14:
@NycuRO I do, by reverting this I can join my server again without crashing(Synapse).

NycuRO:
I don't have this issue man. What error you have?

File14:
There is no error.

PR #1188, opened 2016-10-23
Remove resticts name. Wtf, why exist this?! 

### LINKED ISSUE #1181: Remove plugin restricted names.
Just a simple suggestion, why Nukkit needs to block plugins that contains specific names?

https://github.com/Nukkit/Nukkit/blob/master/src/main/java/cn/nukkit/plugin/PluginManager.java#L172

It would be better if that was removed or at least changed to "equalsIgnoreCase".

And yes, I know PocketMine-MP also blocks those same names, but why block it?

NycuRO:
Fix #1181 .

Why exist this? E.g. plugin: MinecraftMarket . This works with Pokkit. Why is restricted plugin?

### DIFF
--- a/src/main/java/cn/nukkit/plugin/PluginManager.java
+++ b/src/main/java/cn/nukkit/plugin/PluginManager.java
@@ -169,12 +169,6 @@ public class PluginManager {
                         PluginDescription description = loader.getPluginDescription(file);
                         if (description != null) {
                             String name = description.getName();
-                            if (name.toLowerCase().contains("nukkit") || name.toLowerCase().contains("minecraft") || name.toLowerCase().contains("mojang")) {
-                                this.server.getLogger().error(this.server.getLanguage().translateString("nukkit.plugin.loadError", new String[]{name, "%nukkit.plugin.restrictedName"}));
-                                continue;
-                            } else if (name.contains(" ")) {
-                                this.server.getLogger().warning(this.server.getLanguage().translateString("nukkit.plugin.spacesDiscouraged", name));
-                            }
 
                             if (plugins.containsKey(name) || this.getPlugin(name) != null) {
                                 this.server.getLogger().error(this.server.getLanguage().translateString("nukkit.plugin.duplicateError", name));


Muqsit:
Maybe to prevent glitches such as the 'null' username bug that existed.

NycuRO:
@Muqsit Neh. More usefull without

NycuRO:
And wtf ,not exist plugin with null.

MrPowerGamerBR:
@Muqsit this affects plugin names, not player names.

Muqsit:
Isn't it a good thing? It can avoid present/future code conflicts.

NycuRO:
@Muqsit Is good without. Wtf man ,isn't player name. Is plugin name.

Not exist plugin without name == null man. Wtf. Why restrict?

MrPowerGamerBR:
@Muqsit why it would conflict? It is the plugin name.

If I create a plugin called "Minecraft", nothing will explode, it is just a restriction created in PM-MP for no reason.

Maybe exact names should be restricted, but why should restrict plugin that contains those names?

Example: If I create a plugin called "MinecraftRevolution", Nukkit blocks it because it contains the word "Minecraft".

Muqsit:
It's better to be left as it is than to argue over it I guess. Let someone Luke Pub4Game or MagicDroid share their opinion. :~

Snake1999:
:) 
For my personal opinion, names like "MojangPlugin" violated the name rights of mojang company. As some of our developers could push plugins into commercial uses, it could be our duty to protect name rights for mojang, for our projects won't have a DMCA takedown.

boy0001:
@Snake1999 
Actually, but enforcing certain plugin names it indicates that you are responsible for what plugins people use. I would think you'd much rather distance yourself and have whoever is running the server responsible for whatever they want to use Nukkit for. 

If you take it upon yourself to protect the rights of Mojang, it'll just make things more difficult than having the server operator responsible.

NycuRO:
@Snake1999 Wtf man. Add "Mojang" name not , Minecraft or other... In first time exist MinecraftMarket plugin :/

fromgate:
About Minecraft and Mojang: Guys, you just need to read EULA.

You can't use the word "Minecraft" in your product title if it placed in first position.
So if you name your plugin "MinecraftMarket" you break the EULA. But you can name it "Market for Minecraft" and this will be ok. 

And I think same for the word "Mojang".

But what about Nukkit? I think Nukkit team could choose any rules of "trademark usage". But I think there's no reason to forbid usage of the word "Nukkit", because lot of plugins will transfer from other system. And normal solution is to add word "Nukkit" to determine that this plugins is created for Nukkit. For example: NukkitLaser, NukkitReactions, NukkitPEX etc.

NycuRO:
@fromgate MinecraftMarket isn't my plugin. Is plugin comercial for PC. With pokkit,this plugin work: https://www.minecraftmarket.com/

PR #1193, opened 2016-10-24
0.16 Update (Part1, 2, 3, 4....) Now we can join the pure server. Have a lot of bugs. Hope you to fix them.

boybook:
0.16 Update (Part1, 2, 3, 4....) based of https://github.com/yescallop/Nukkit/tree/0.16

Now we can join the pure server.
Have a lot of bugs:
1. Client may stuck when joining the server (Because of the map? the EntityData?).
2. Client will crash about UpdateBlockPacket about BinaryStream of SignedVarInt/SignedVarLong
3. Client sprint not works good.
4. The player list appears two identical our own name.
5. Abnormal bubbles show.
6. Incorrect json of command list (new in 0.16).
.......
Maybe because of the new VarInt/VarLong in Binary class.


boybook:
[ERROR] /home/ubuntu/Nukkit/src/main/java/cn/nukkit/utils/Binary.java:[5,33] package com.sun.javaws.exceptions does not exist
I will fix it tomorrow.

MrPowerGamerBR:
@boybook you shouldn't use com.sun.* packages, if you create something with the "com.sun.*" packages, you will only be able to run it with Oracle's JDK (OpenJDK will fail to run because of the missing packages)

Probably CircleCI/TravisCI uses OpenJDK, that's why it fails.

MrPowerGamerBR:
Forget what I said, just remove the ```import com.sun.javaws.exceptions.InvalidArgumentException;``` and it magically works.

I compiled and the client connects without any issues, so... 👍

boybook:
Sorry about it, I forgot removing the useless import...

Pub4Game:
@boybook great, but it should be do in 0.16.0 branch. @Sanke1999 I think you need revert it and create new branch "0.16.0" with this changers.

PR #1203, opened 2016-10-26
Better splash potion effect

Creeperface01:


Pub4Game:
From this moment on, write a full description of the changes, otherwise, I will just close the pull request

Creeperface01:
Its hard :D but i fixed splash area size + potion duration is decreased by distance between potion and player

File14:
Thanks, it was so weird before!

PR #1205, opened 2016-10-28
Validate horizontal and vertical movement separately

### LINKED ISSUE #420: Change how the "invalid movement" works.

boy0001:
- Vertical movement was already checked above: if (diffY >= -yS || diffY
<= yS)
- Fixes #420 

Alternative solution to #1128

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -1248,9 +1248,9 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
             }
 
             if (diffX != 0 || diffY != 0 || diffZ != 0) {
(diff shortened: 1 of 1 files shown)


MagicDroidX:
#1128 We are trying to find a solution for stairs and stabs and I wonder if this pull has made a solution?

boy0001:
@MagicDroidX 
#1128 would let you run fast if you are adjacent to slabs or stairs regardless of whether you are climbing them. It's also slow to check all the adjacent blocks each player move. The correct solution is to limit the vertical and horizontal movement separately.

PR #1208, opened 2016-10-29
Add block place sound

Creeperface01:
@Pub4Game Adding sound when player places a block

BTW why there still are already merged commits? o.O


MagicDroidX:
This may have to be reworked as we have updated to 0.16 ;D

Issue #1218, opened 2016-10-30
Error when using /give


MrPowerGamerBR:
### Issue description

I was trying to use /give and... well, errors related to JSON.
### Steps to reproduce the issue

1. Join a server
2. Use /give yournickname 1 1
### OS and versions

- Nukkit: Este servidor estß rodando Nukkit 1.0dev ???(Apple)?(Pie)? implementando versÒo da API 1.0.0 para Minecraft: PE v0.16.0 alpha (versÒo do protocolo 91)
- OS: Windows 10
- Device: Windows 10
### Crashdump, backtrace or other files

```
2016-10-30 [ALERT] com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected BEGIN_OBJECT but was NUMBER
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:221)
        at com.google.gson.Gson.fromJson(Gson.java:814)
        at com.google.gson.Gson.fromJson(Gson.java:879)
        at com.google.gson.Gson.fromJson(Gson.java:852)
        at cn.nukkit.Player.handleDataPacket(Player.java:2763)
        at cn.nukkit.network.Network.processPackets(Network.java:186)
        at cn.nukkit.network.Network.processBatch(Network.java:157)
        at cn.nukkit.Player.handleDataPacket(Player.java:1832)
        at cn.nukkit.network.RakNetInterface.handleEncapsulated(RakNetInterface.java:154)
        at cn.nukkit.raknet.server.ServerHandler.handlePacket(ServerHandler.java:123)
        at cn.nukkit.network.RakNetInterface.process(RakNetInterface.java:65)
        at cn.nukkit.network.Network.processInterfaces(Network.java:76)
        at cn.nukkit.Server.tick(Server.java:981)
        at cn.nukkit.Server.tickProcessor(Server.java:788)
        at cn.nukkit.Server.start(Server.java:767)
        at cn.nukkit.Server.<init>(Server.java:458)
        at cn.nukkit.Nukkit.main(Nukkit.java:65)
Caused by: java.lang.IllegalStateException: Expected BEGIN_OBJECT but was NUMBER
        at com.google.gson.internal.bind.JsonTreeReader.expect(JsonTreeReader.java:139)
        at com.google.gson.internal.bind.JsonTreeReader.beginObject(JsonTreeReader.java:70)
[...]

Issue #1220, opened 2016-10-30
Command doesn't work properly.


EnderKenji:
### Issue description

Commands not working properly.

 It only reads 1 ars, like /mw tp 1, but only read /mw tp and it does nothing. If i do /mw tp only, it send me usage of the command, detect args[1] as null.  You also can try /teleport command. 

but, command works fine if u send command as ConsoleCommandSender / on console.
### Steps to reproduce the issue

Already told in desc.
### OS and versions

- Nukkit: Nukkit 1.0dev 「蘋果(Apple)派(Pie)」 implementing API version 1.0.0 for Minecraft: PE v0.16.0 alpha (protocol version 91)
- OS: Windows 10
- Device: Windows 10
##

boybook:
I'm recoding the command encode, it will be fixed in a moment.

PR #1224, opened 2016-10-31
Rewrote the command args decode and recheck all default commands.

boybook:
- Rewrote the command args decode and recheck all default commands.
- Fix the crash when opening the chest.


boybook:
Fixed https://github.com/Nukkit/Nukkit/issues/1218 and https://github.com/Nukkit/Nukkit/issues/1220

Issue #1222, opened 2016-10-30
server doesn't stop


Creeperface01:
### Issue description

sometimes server doesn't stop properly.
There is this message in the console "Took too long to stop, server was killed forcefully!" but process is still running.. so server won't be started again.
### Steps to reproduce the issue

happens randomly
### OS and versions
- Nukkit: 1.0-SNAPSHOT
- OS: windows 10
- Device: HP
### Crashdump, backtrace or other files

> 20:21:45 [DEBUG] Disabling all plugins
> 
> 20:21:45 [INFO] Disabling Annihilation v1.0
> 
> 20:21:45 [INFO] [Annihilation] Annihilation disabled
> 
> 20:21:45 [INFO] Disabling GTCore v1.0
> 
> Took too long to stop, server was killed forcefully!
> - ...

Creeperface01:
and i can still send commands after this

MagicDroidX:
@boy0001 Maybe this is due to the new logger?

Creeperface01:
No it happened before too

Creeperface01:
Or when was new logger added?

boy0001:
About a day before you created this issue.

Creeperface01:
So its not the cause

Creeperface01:
But it happened again today, server stuck while stopping command reader thread, but wasnt stopped Only message appeared in the console

boybook:
It never happened on my computer. Mayne your plugins caused it?

Creeperface01:
Plugins can cause, commandReader cannot be interupted?

boy0001:
@Creeperface01 
Nukkit does the following on shutdown
 - Disable all plugins
 - Unload levels
 - Remove event handlers
 - Stopping tasks
 - Closing console
 - Stopping network interface
 - Disabling timings
 - Stopping threads

According to the log you posted above, it's getting stuck on disabling GTCore. Do you have the log of it getting stuck on any other stage?

Creeperface01:
Yes but server killer should stop server shouldnt it?

MagicDroidX:
There might be a running thread created by a plugin which prevent to server from being stopped.

Creeperface01:
But i think it happened even without plugins.  In PocketMine it happens too sometimes

Creeperface01:
But i have Only mysql query thread

PR #1234, opened 2016-11-01
halt instead of system.exit

boy0001:
Should always stop the server where exit fails
#1222

### DIFF
--- a/src/main/java/cn/nukkit/utils/ServerKiller.java
+++ b/src/main/java/cn/nukkit/utils/ServerKiller.java
@@ -25,7 +25,7 @@ public class ServerKiller extends Thread {
             // ignore
         }
         System.out.println("\nTook too long to stop, server was killed forcefully!\n");
-        System.exit(1);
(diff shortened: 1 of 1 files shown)


MagicDroidX:
There are some shutdown hook methods in Nukkit.

boy0001:
@MagicDroidX 
The only hook is [this](https://github.com/Nukkit/Nukkit/blob/master/src/main/java/cn/nukkit/raknet/server/RakNetServer.java#L98) which prints the message `RakNet crashed!` if the server isn't shutdown. 

Since the server is shut down in this specific instance, that message would never be printed.

Creeperface01:
@MagicDroidX sometimes server isn't fully stopped, and command reader thread stays running

Creeperface01:
but @boy0001 's method doesn't fix it..

Creeperface01:
@magicdroidx so how to solve this? Can daemon thread fix it it?

Issue #1280, opened 2016-11-16
EXPERIENCE fixed source


OverTook:
There is a mistake about calculateRequireExperience. (Sorry, i don't know how to use Pull repuests)

if you use /xp command when your level is 15, it skip 16 // and, i fix it. please merge it!
(sorry for my bad enlgish.)

~~~~~~~~~~~~~~~~~~~
    public static int calculateRequireExperience(int level) {
        if (level < 16) {
            return 2 * level + 7;
        } else if (level >= 16 && level <= 31) {
           //Fixed 17 to 16;
            return 5 * level - 38;
        } else {
            return 9 * level - 158;
        }
    }

Snake1999:
Create a pull request

OverTook:
@Snake1999 I said "Sorry, i don't know how to use Pull repuests"

PR #1281, opened 2016-11-16
fixed "calculateRequireExperience(int level)"

OverTook:
it will fix an error when use /xp command
(now i know how to create pull requests! :D)

https://github.com/Nukkit/Nukkit/issues/1280


File14:
Thanks, I noticed this glitch

PR #1284, opened 2016-11-17
calculate BB from scale

Creeperface01:


Snake1999:
Great!
I'll wait for hear from our stuff. I'm going to NOI contest(National Olympiad of Informatics) in China, I'll have a break after the 3-day contest is over

PR #1293, opened 2016-11-25
Faster getBlock() method

Creeperface01:


Creeperface01:
i think, its pointless to access block cache twice

yescallop:
Seems It doesn't access block cache twice

Creeperface01:
So its faster, isnt už?

Creeperface01:
Yes it does, because, it checks if it contains (first access) and then get (Second)

Creeperface01:
getMethod: getNode(hash(key), key)
contains method: getNode(hash(key), key) != null

absolutely same

Creeperface01:
@MagicDroidX i think it can be merged

Pub4Game:
Mmm... Cool, thanks, so little conflict... You can be patient and do not merge a pull request, please?

PR #1301, opened 2016-11-28
Fix adventure settings not updated for GUI gamemode changes - PocketMine-MP

File14:
Fix adventure settings not updated for GUI gamemode changes from PocketMine-MP!


File14:
Wtf, why did my commits just randomly got added? I never wanted them to be added into this PR. ._.

PR #1326, opened 2016-12-17
Make the Level flag work so that the beacons light up. No GUI yet.

daveboden:
Please review; feedback welcome.


File14:
Thanks, good for building

Issue #1346, opened 2016-12-30
Bucket disappears picking up liquids


qazedc1212:
Write a short description about the issue
When trying to pick up liquids, bucket disappears, placing liquids work though
### Steps to reproduce the issue

1. Place water/lava
2. Try to pick it up
3. Bucket disappears, and liquid stays

### OS and versions
* Nukkit Version: 1.0DEV
 
* Java Version: 1.8.0_111

* Host OS: Ubuntu 14.04

* Host Configuration: 1 GB RAM, 20 GB SSD drive, 2 core CPU from Intel, 150 Mb internet upload

* Client Edition: Win 10/iOS

* Client Version: 1.0.0

PR #1349, opened 2016-12-31
Enchant table craft fix & bucket fix

imjack:
I'm not too good with git as you can probably tell. This is my last try at adding the fix.


Pub4Game:
Formatting is terrible, I do not understand what you fixed with buckets

imjack:
Formatting is the standard eclipse I can't fix what I don't understand. Buckets disappear when used. #1346

Creeperface01:
nukkit uses IntelliJ formating probably :D

imjack:
Well if that's the case at least state what you'd like to be used.

Creeperface01:
but i think a bit longer lines should be still in one line, especially method's head

Pub4Game:
@imjack Please fix the code formatting. Then I'll add your changes.

PR #1356, opened 2017-01-01
fix command time like singleplayer values

NycuRO:


XShockinFireX:
@NycuRO Why not edit the constants?

NycuRO:
@XShockinFireX What are you doing?

MagicDroidX:
Why not change the constants' values instead?

PR #1357, opened 2017-01-04
fix duplicate player list

jhelom:


Pub4Game:
I think checking the UUID will be better :)

Creeperface01:
why? checking UUID will be slower isn't it?

Pub4Game:
UUID will be 100% unique

jhelom:
https://github.com/Nukkit/Nukkit/pull/1358

PR #1358, opened 2017-01-05
fix duplicate player name in player list. changed comparison from ID to UUID

jhelom:


jhelom:
https://github.com/Nukkit/Nukkit/pull/1357

PR #1360, opened 2017-01-06
Add VehicleExitEvent

virustotalop:
Added VehicleExitEvent for player. I didn't see any way to get when a player or any entity for that matter exits a vehicle. If this is already in the API you can just close the pull request but I didn't see any way to do it.


MagicDroidX:
Conflicted :[

PR #1364, opened 2017-01-07
Add A Temporary Ban Command

angusbeefsteak:
This is the start of the /tempban command

This will create a temporary ban for the player specified
Usage:  /tempban [Player name] [Ban Length] [Ban reason]

Example:  /tempban PiggyPlauge87 30Days Fly Hacking


Pub4Game:
This command implemented by the plugin.

PR #1367, opened 2017-01-08
Fix double plant drops

Creeperface01:


Creeperface01:
WTF, why there are already merged commits? :D

Issue #1359, opened 2017-01-05
Chunks bug


vzlomka76:
is such a hole has appeared in the world, where it came from - no one knows
Notify me of it the player, if it enter, then you will suffocate in the wall, but you can go through it, through worldedit (plugin Regions) I tried to close it, but all the blocks disappear. What should I do?
![2017-01-05 12-03-16](https://cloud.githubusercontent.com/assets/16573760/21675264/42de01ee-d342-11e6-9bfc-19067bd5c27f.png)

qazedc1212:
Try placing a block in the broken chunk. The block should be replaced by what was previously there

vzlomka76:
@qazedc1212 if I put a block, then it is put, but if i death or quit it disappears

Pub4Game:
@yescallop Any ideas?

yescallop:
@Pub4Game I think all chunk sections should be sent, doing this with inflating hardly affects the performance. And it could fix another bug of teleporting

Pub4Game:
@yescallop You try to fix it?

yescallop:
Is this bug in 1.0.0?

Pub4Game:
Yes

kvetinac97:
Does it have something in common with the PC world "chunk errors" (some regions not loading)?

vzlomka76:
@kvetinac97 Map generated in nukkit

Creeperface01:
the similar bug appears if you are in normal world and then teleport to another empty world, you still see chunk sections from last world

Pub4Game:
This problem was solved in PM-MP. Someone will do it?

Creeperface01:
in nukkit it isn't async?

vzlomka76:
@Creeperface01 at first everything was fine, I saw this place, where the river flowed, but then one day, in many places there were holes, new worlds I did not create (there is only one world)

kvetinac97:
Strange bug :/ Changed the map which bugged to first map, but it still lags... maybe Nukkit isn't loading so many chunks, or uses different ordering...?

Creeperface01:
@kvetinac97 měl jsem podobný problém ale pouze při prvním testu teď to asi ani nedělá každopádně zdá se mi to nějak více vytížené :D

xpyctum:
#326 copy (maybe, 50%50)

Issue #1268, opened 2016-11-07
Teleport issue?


Creeperface01:
### Issue description
players gets stuck in air on teleport, usually between worlds (maybe long distance). THis doesn't happen always, but very often

### Steps to reproduce the issue
Teleport between worlds

### OS and versions

* Nukkit Version: 1.0-SNAPSHOT
 
* Java Version: 8

* Host OS: linux

* Host Configuration: Intel(R) Xeon(R) CPU E3-1271 v3 @ 3.60GHz, 2 GB RAM, 100 Mb

* Client Edition: all

* Client Version: 0.16.1

Snake1999:
How do you 'teleport between worlds'? By command /tp?

If you did it by plugin, PASTE YOUR PLUGIN CODE, if you don't have the code, ask plugin author.

Creeperface01:
player.teleport()

File14:
Same issue.

boybook:
Chunk packets loss?

Creeperface01:
@boybook maybe nukkit doesn't send all chunk sections, (e.g. empty sections)

PR #1387, opened 2017-01-18
Fix teleport issue between worlds

yescallop:
Send all chunk sections to player
Closes https://github.com/Nukkit/Nukkit/issues/1268
Maybe closes https://github.com/Nukkit/Nukkit/issues/1359


Pub4Game:
Hmm... I'll test this today

PR #1389, opened 2017-01-19
Timings refactor

MrPowerGamerBR:
Okay, this isn't really a fix, it is actually a refactoring in the Timings v2 related packages.

Timings was refactored from ```cn.nukkit.timings``` to ```co.aikar.timings```, this WILL break plugins that uses the Timings v2 API, but as far as I know there isn't any public plugins using those APIs, and, even if it does break, the fix is just a package rename, so it is **that** breaking.

Changed the user agent to Nukkit instead of Sponge because Nukkit is now allowed in @aikar's Timings v2 website.

Added the original @aikar's file headers and removed the @author headers for classes from @aikar.

I think that's it, at least now Nukkit uses the correct Timings v2 user agent instead of borrowing the user agent from Sponge 👍


MagicDroidX:
Please deal with the conflicts so that it can be merged. ;P

MrPowerGamerBR:
@MagicDroidX well, I will need to create a new pull request since I deleted the fork I made those changes.

So I will create a new pull request with the changes

MrPowerGamerBR:
@MagicDroidX #1432

PR #1401, opened 2017-01-24
Improve the hack of empty subchunks

yescallop:
Performance up up up!
However, there is still some subchunks are not completely cleared.
#blameMojang https://bugs.mojang.com/browse/MCPE-19818


Creeperface01:
client performance or server performance?

PR #1406, opened 2017-01-27
Implement "notifyACK" method, change AvailableCommandsPacket to use "needsACK", fixes #1378

### LINKED ISSUE #1378: Sometimes all commands are unknown
### Issue description

Write a short description about the issue
-Sometimes all commands are unknown

-There is no step,just when server is running,and some of the players enter the command is unknown command.
-Sometimes getting out of the server can not only solve the problem,but sometimes not.
-Restart the server can solve this problem.
!!!This problem is only for some players,other players are normal.

### OS and versions

* Nukkit Version: Nukkit:1.0dev  API:1.0.0  protocol:100

 
* Java Version: 1.8.0_111

* Host OS: Linux

* Host Configuration: 4GB RAM, 35GB SSD drive, 4 core CPU from Intel, 5 Mb internet upload

* Client Edition: Windows 10 , Android

* Client Version: 1.0.0

### Crashdump, backtrace or other files

* Nothing
```
<!--- Copy and paste your backtrace here... -->
```Nothing
* ...

MrPowerGamerBR:
For some reason, while Nukkit processes the ACK notification, it doesn't do anything with it, so yes, any packet that were using "notifyACK" couldn't do anything with the notification, after all it wouldn't never receive it.

This commit changes that, it implements a very very very basic notifyACK method, changing the player's "needsACK" map (which wasn't used for NOTHING before this commit except adding packets that used the "needsACK" boolean (but it would just stay there with a "false" value so...)) to true if the server receives a notification from the client showing that yes, the client received the packet successfully.

Why that matters? This allows us to know IF the packet reached the client, allowing us to do a workaround for the #1378 issue.

So now we send a packet saying "hey client, when you receive the AvailableCommandsPacket packet, could you send a notification to the server showing that yes, you received it?", then, after sending, a thread is created (so we don't hang the main thread) and, after 3 seconds, if we DIDN'T receive the client's notification about receiving the AvailableCommandsPacket, we try sending again the packet until the player successfully receives it. (or if he logs out, whatever comes first)

Also a note: I couldn't test it on a very populated server, so... can someone test this commit on a public server before pushing to master? I tested on my local server and it worked, but I didn't have a chance to test on a public server.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -497,7 +497,23 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
         if (count > 0) {
             //TODO: structure checking
             pk.commands = new Gson().toJson(data);
-            this.dataPacket(pk);
+            int identifier = this.dataPacket(pk, true); // We *need* ACK so we can be sure that the client received the packet or not
+            Thread t = new Thread() {
+                public void run() {
+                    while (true) {
+                        // We are going to wait 3 seconds, if after 3 seconds we didn't receive a reply from the client, resend the packet.
+                        try {
+                            Thread.sleep(3000);
+                            boolean status = needACK.get(identifier);
+                            if (!status && isOnline()) {
+                                sendCommandData();
+                                return;
+                            }
+                        } catch (InterruptedException e) {}
+                    }
+                }
+            };
+            t.start();
         }
     }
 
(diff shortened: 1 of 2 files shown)


MrPowerGamerBR:
Also a fun fact: Only now I figured out that "ACK" means "acknowledgment", now it makes sense why it is called "ACK" response... wow...

Snake1999:
This might be the way to fix. Testing...

DaMatrix:
Can confirm that this works.

MrPowerGamerBR:
Also, the "needsACK" map should be replaced with a expiring cache from Guava, if kept this way, the packets will stay on the needsACK until the player quits the server, causing unnecessary memory use.

MrPowerGamerBR:
Wait a minute... I found a little issue in the code...

PR #1408, opened 2017-01-27
MobHeads are stackable to 64

MrPowerGamerBR:
From https://github.com/pmmp/PocketMine-MP/commit/c36fc8c0272e2a01728b1c95fbbb77dea515757f


Snake1999:
Thank you.
BTW, we should follow **the mojang client** instead of what other software(such as ~~PMMP~~) says. Only after the **experiments in mojang client** says the mob head should be stacked to 64, should we do this modification.

PR #1411, opened 2017-01-28
Remove "while (true) { ... }" from sendCommandData()

MrPowerGamerBR:
I was using that when I was debugging the ACK response, but I forgot to remove it when I submitted the pull request.


MrPowerGamerBR:
My branch is outdated, let me do that again...

PR #1250, opened 2016-11-03
The pull request that changed Nukkit generator forever!

Niall7459:
This is a very very big step foward in our generator!

Thanks to @Creeperface01  and @NycuRO I have been able to port their nukkit plugin (BiomeMaster) to the nukkit core.
https://github.com/NycuRO/BiomeMaster
This PR should be ready to merge but there may be a few unfinished things.

**Adds:**
-Jungle biome
-Roofed forest biomes
-Savanna biome
-Mushroom islands

-New trees for biomes

**TODO:**
- Check the other biomes are appearing (very hard to find some of them)
- Add the trees to generate when planting sapling.
**- Possible change to fit different structure?**
Screenshot:
screen shot 2016-11-03 at 16 17 52
screen shot 2016-11-03 at 16 42 59

PR #1416, opened 2017-01-28
CustomBiomes Plugin Built in!

yungtechboy1:
Extends #1250 which was abandoned...

Tested and Custom Biomes work!
More biomes to come  >:)


Creeperface01:
I would merge it :D

PR #1420, opened 2017-01-29
Allow Custom Enchants and Effects!

yungtechboy1:
Since Block and Item lists are public and allow for easy implementation of Custom Items and Blocks. I made Enchants and Effects public to allow custom enchants and effects.

I will be releasing a Custom Enchant Plugin soon XD
This will help a lot.

### DIFF
--- a/src/main/java/cn/nukkit/item/enchantment/Enchantment.java
+++ b/src/main/java/cn/nukkit/item/enchantment/Enchantment.java
@@ -25,7 +25,7 @@ import java.util.concurrent.ThreadLocalRandom;
  */
 public abstract class Enchantment implements Cloneable {
 
-    protected static Enchantment[] enchantments;
+    public static Enchantment[] enchantments;
 
     public static final int ID_PROTECTION_ALL = 0;
     public static final int ID_PROTECTION_FIRE = 1;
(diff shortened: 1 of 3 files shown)


NycuRO:
No. Because you can hack items with that.

yungtechboy1:
@NycuRO that ONLY if the plugin Allows it

yungtechboy1:
This would not cause a network issue or exploit. Only a plugin dev can add item to this. No one from the network can add a New Enchant or anything like that. And if that was the Case then Item.list and Block.list should be Protected as well!

Snake1999:
No no no. Listen to events instead.

MrPowerGamerBR:
I think it would be a good idea, but a proper API should be added instead of changing the variables to public

yungtechboy1:
@Snake1999 
It cant be an Event Listener! Because it is harder to calculate things Pre and Post-Attack. The current API for enchantment would allow this perfectly. I have already built my custom Enchant plugin around this. This pull isnt necessarily for me and my plugin its for other who want to make custom enchants. Plus in the plugin I would have to wait for the interact Event or Attack Event then Recalculate everything over again
@Snake1999

PR #1431, opened 2017-01-30
Synchronize console commands, warn when a plugin tries to run a command in a async task, fixes #1248

### LINKED ISSUE #1248: Console commands are async, and not safe
Console commands are executed from the CommandReader thread rather than the main thread:
 - https://github.com/Nukkit/Nukkit/blob/master/src/main/java/cn/nukkit/command/CommandReader.java#L70

For example `save-all` results in modifications to the RandomAccessFile from the CommandReader thread. If an autosave is occuring at the same time it'll corrupt the world.

Possible solutions:
 - Schedule the command to run on the main thread
 - Use locks to prevent concurrent modification to the file (#1102)

MrPowerGamerBR:
Also adds the ```Server#isPrimaryThread()``` to check if the current thread is the primary thread or not.

If a plugin tries to run a command in a async task, it warns the that a plugin is running a command in a async task. (Which is not good, see #1248)

### DIFF
--- a/src/main/java/cn/nukkit/Server.java
+++ b/src/main/java/cn/nukkit/Server.java
@@ -186,7 +186,10 @@ public class Server {
 
     private Level defaultLevel = null;
 
+    private Thread currentThread;
+    
     public Server(MainLogger logger, final String filePath, String dataPath, String pluginPath) {
+        currentThread = Thread.currentThread(); // Saves the current thread instance as a reference, used in Server#isPrimaryThread()
         instance = this;
         this.logger = logger;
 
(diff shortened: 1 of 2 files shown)


MrPowerGamerBR:
Also, if a command is dispatched async, the command will NOT be cancelled, it will be executed normally, but it will print a warning in the console.

MrPowerGamerBR:
Also I should mention that, of course, it won't warn the console when running commands in the console, I just made it throw an error just as an example, the pull request won't cause that. (Unless if a plugin is DISPATCHING a command in a async task)

MrPowerGamerBR:
So doing this
```
        Server.getInstance().getScheduler().scheduleTask(new Runnable() {
            public void run() {
                // This is bad!
                Server.getInstance().dispatchCommand(sender, "list");
            }
        }, true);
```
Will trigger an async warning, but this:
```
        Server.getInstance().getScheduler().scheduleTask(new Runnable() {
            public void run() {
                // This is bad!
                Server.getInstance().dispatchCommand(sender, "list");
            }
        });
```
Won't trigger an warning

MrPowerGamerBR:
@MagicDroidX don't forget to close https://github.com/Nukkit/Nukkit/issues/1248 ;)

MagicDroidX:
👍

Issue #1325, opened 2016-12-17
Terminal command error


DaMatrix:
### Issue description

Sometimes, when entering a command in the terminal, I get the following stack trace:
`Exception in thread "Console" java.lang.StringIndexOutOfBoundsException: String index out of range: 2
        at java.lang.AbstractStringBuilder.charAt(AbstractStringBuilder.java:237)
        at java.lang.StringBuilder.charAt(StringBuilder.java:76)
        at jline.console.ConsoleReader.wcwidth(ConsoleReader.java:545)
        at jline.console.ConsoleReader.backspace(ConsoleReader.java:1050)
        at jline.console.ConsoleReader.backspaceAll(ConsoleReader.java:1036)
        at jline.console.ConsoleReader.resetPromptLine(ConsoleReader.java:3859)
        at cn.nukkit.command.CommandReader.readLine(CommandReader.java:45)
        at cn.nukkit.command.CommandReader.run(CommandReader.java:60)`
. After this my terminal is completely unusable, both from SSH and direct keyboard usage. It can only be fixed by Ctrl+C-ing Nukkit, which is less than ideal.

### Steps to reproduce the issue

1. Start nukkit
2. Spam any command into the terminal

### OS and versions

* Nukkit Version: 5eb62df76a024f415cd2330b425bd75949d861d8
 
* Java Version: Reproduced on latest openjdk-7 and openjdk-8

* Host OS: Ubuntu 16.04 (Server)

* Host Configuration: Reproduced on an overpowered dedi, a ThinkPad T420, a MacBook pro, and a 4Gb VPS.

* Client Edition: Doesn't matter, can even happen with nobody online.

* Client Version: 0.16.1

AlainBecker:
I've seen this as well, running on an ARM server. 

At first glance, this shouldn't result in any server dysfunction, so would rather be fit for the "cosmetical" issue category...

Upgrading the `jline` dependency  to [the latest upstream 2.x release](https://github.com/jline/jline2) (or even [3.x](https://github.com/jline/jline3) ?) might be enough to solve this.

MrPowerGamerBR:
Maybe fixed in https://github.com/Nukkit/Nukkit/commit/23b035e7f0659434b44286c0554947ef3f9f67f8 because I changed the readLine() method to be something more simple and pretty. (and that's how JLine recommends)

DaMatrix:
OK, I'll give it a try. 

-------- Original message --------

[...]

PR #1442, opened 2017-01-31
Upgrade jline to 2.14.2

MrPowerGamerBR:
May fix https://github.com/Nukkit/Nukkit/issues/1325

No issues, but a upgrade is always nice, right?

**TODO:** Upgrade to JLine 3

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,9 +30,9 @@
             <version>1.16</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


MrPowerGamerBR:
To be honest I think it is a issue with Nukkit's console reader, not JLine... So I will investigate a little bit before opening this again.

PR #1447, opened 2017-01-31
Add TNT explosion sound

MrPowerGamerBR:


MrPowerGamerBR:
That's why I should never use GitHub's file editor, I always forget something when copying code from Eclipse to there. (I used the file editor because I didn't have a up to date fork on my PC)

PR #1452, opened 2017-02-01
Disable titleTick()

MrPowerGamerBR:
Okay, so now I could reproduce this issue again.

Disabling titleTick() fixes this, I will take a look on how to fix the titleTick() later, but for now, it is better disabling it. (if kept enabled, it will flood the console like in that screenshot)

### DIFF
--- a/src/main/java/cn/nukkit/Server.java
+++ b/src/main/java/cn/nukkit/Server.java
@@ -1079,8 +1079,9 @@ public class Server {
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


PocketHDJ:
(ー_ー)!!wtf

MrPowerGamerBR:
Also, after fixing the issue

The only issue is... well, now the title doesn't change anymore, but at least doesn't flood the console.

PR #1453, opened 2017-02-01
BossBar API

MrPowerGamerBR:
Adds a simple BossBar API.

You can create BossBar with ```Player#createBossBar("Hello World!", 100);```, this will return a bossBarId, store it for further use.

You can also update BossBar with ```Player#updateBossBar("New Text", 50, bossBarId);```, this will change the BossBar to show "New Text" and it will be half full.

And you can delete it with ```Player#deleteBossBar(bossBarId);```.

**TODO:** Nukkit is missing the Attributes value in AddEntityPacket, however sending the Attributes in the AddEntityPacket causes issues, client bug? (So, while this isn't fixed, the code uses UpdateAttributesPacket)

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -4362,7 +4362,6 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
         attr.setValue(length); // Entity health
         pkAttributes.entries = new Attribute[] { attr };
         this.dataPacket(pkAttributes);
-        this.createBossBar("wow", 100);
         // And then the boss bar text
         SetEntityDataPacket pkMetadata = new SetEntityDataPacket();
         pkMetadata.eid = bossBarId;
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -20,6 +20,7 @@ import cn.nukkit.entity.EntityHuman;
 import cn.nukkit.entity.EntityLiving;
 import cn.nukkit.entity.data.*;
 import cn.nukkit.entity.item.*;
+import cn.nukkit.entity.mob.EntityCreeper;
 import cn.nukkit.entity.projectile.*;
 import cn.nukkit.event.block.ItemFrameDropItemEvent;
 import cn.nukkit.event.block.SignChangeEvent;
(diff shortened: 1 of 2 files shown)


MrPowerGamerBR:
*facepalm* I kept a small piece of code when I was debugging the API, do not merge yet.

MrPowerGamerBR:
Done!

I don't know why that was here, but now the API works 100% (well... it worked before... but every time you updated the bossbar it would create another bossbar...)

MrPowerGamerBR:
And here is a screenshot

Creeperface01:
but we can not use custom health

MrPowerGamerBR:
@Creeperface01 I hardcoded it to 100 because... well, do you really need more precision than 100%?

(Unless if you are talking about another thing... then I don't know, I can use the Max Health attribute just fine)

Creeperface01:
Maybe replace length with percent

MrPowerGamerBR:
@Creeperface01 well, length is the bar percentage, I just couldn't figure out a good word for that, maybe I will change that variable later 😉

SuperMaXAleX:
I recommended to add the possibility to custom the max health, because, for example, if you want a timer for 1200 seconds and you want to merge it on the boss bar, you will need to change the max health. :/

Creeperface01:
You dont have to

MrPowerGamerBR:
@SuperMaXAleX I will add that later. 👍

SuperMaXAleX:
@MrPowerGamerBR Thanks you very much!

MrPowerGamerBR:
Also, there is a small issue with this API: If you set the health to 0, the bossbar sometimes vanishes. (probably a client side issue)

Creeperface01:
Because entity with 0 health is death..

Snake1999:
Well done :) 
Will it better if BossBar is wrapped into a special class (like PlayerFood), and an API added for changing boss bar health(that purple colored bar) ?)

MrPowerGamerBR:
@Snake1999 I still think that adding everything in separate classes is not a very good idea, it makes the code more redundant. (And if it is just an separate class, why bother adding it to Nukkit's source)

Unless if you are saying to create a different class for the BossBar API AND make the Player's BossBar API methods "redirect" to the new class (but not removing the methods from the player class)

Snake1999:
@MrPowerGamerBR 
```
Unless if you are saying to create a different class for the BossBar API AND make the Player's BossBar API methods "redirect" to the new class (but not removing the methods from the player class)
```
That's what i meant... I agree with this idea :)

MrPowerGamerBR:
@Snake1999 I was just asking that because in your 1.0.3.0 branch you actually removed the "transferTo" method from the Player class and just kept the "Transferring" class

PR #1457, opened 2017-02-03
Add NukkitRunnable

MrPowerGamerBR:
Using Nukkit Scheduler is a mess, so I created NukkitRunnable.

It works just like BukkitRunnable... except it is for Nukkit 😉 

Example:
```
        new NukkitRunnable() {
            public void run() {
                for (Player p : Server.getInstance().getOnlinePlayers().values()) {
                    if (bossBars.containsKey(p)) {
                        long bossBar = bossBars.get(p);
                        p.updateBossBar("Hello World! §a§l" + TasselUtils.getRandom().nextInt(0, 800), 100, bossBar);
                    } else {
                        long bossBar = p.createBossBar("Temporary", 100);
                        bossBars.put(p, bossBar);
                    }
                }
            }
        }.runTaskTimer(this, 20, 20);
```

**Why I done this instead of using Nukkit Scheduler?**
Nukkit Scheduler doesn't let you cancel the task inside of the task, making you need to create a lot of unnecessary variables, bloating a lot your code.

Also, it's nice, and it works without any issues.

### DIFF
--- a/src/main/java/cn/nukkit/scheduler/NukkitRunnable.java
+++ b/src/main/java/cn/nukkit/scheduler/NukkitRunnable.java
@@ -20,37 +20,37 @@ public abstract class NukkitRunnable implements Runnable {
 
     public synchronized Runnable runTask(Plugin plugin) throws IllegalArgumentException, IllegalStateException {
         checkState();
-        this.taskHandler = Server.getInstance().getScheduler().scheduleTask(this);
+        this.taskHandler = Server.getInstance().getScheduler().scheduleTask(plugin, this);
         return taskHandler.getTask();
     }
 
     public synchronized Runnable runTaskAsynchronously(Plugin plugin) throws IllegalArgumentException, IllegalStateException  {
         checkState();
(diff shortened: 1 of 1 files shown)


MrPowerGamerBR:
You can use "this.cancel();" inside of an task to cancel it, example:
```
    @EventHandler
    public void onLogin(PlayerJoinEvent e) {
        new NukkitRunnable() {
            public void run() {
                if (/* insert condition here */) {
                    this.cancel();
                    return;
                }
                e.getPlayer().sendTip("§a§lYou logged in! Wow!");
            }
        }.runTaskTimer(m, 0, 20);
    }
```

MrPowerGamerBR:
And yes, you can keep using schedulers if you want to.

MrPowerGamerBR:
Do not pull until https://github.com/Nukkit/Nukkit/issues/1458 is closed

MrPowerGamerBR:
# Pull https://github.com/Nukkit/Nukkit/pull/1459 first!
## The CircleCI/TravisCI test won't pass because this code uses code only available in #1459!

PR #1465, opened 2017-02-06
Better block collision checking

Creeperface01:


Creeperface01:
the physical bounding box (getBoundingBox() ) should be collision bounding box and seconds bounding box should be just bounding box, but that would break plugins etc.. so i named it collision bounding box, but i can change it

PR #1471, opened 2017-02-08
Fixing water holes

alistairwatts:
When STILL_WATER is removed using a bucket we now look for adjacent STILL_WATER and change it to WATER.  This WATER is now able to flow into the space where the STILL_WATER was removed.

The change avoids problems like this:
screenshot_20170208-231519

It looks like the same fix is required for lava, but I haven't checked.  I'll take a look another time.

### DIFF
--- a/src/main/java/cn/nukkit/item/ItemBucket.java
+++ b/src/main/java/cn/nukkit/item/ItemBucket.java
@@ -4,6 +4,7 @@ import cn.nukkit.Player;
 import cn.nukkit.block.Block;
 import cn.nukkit.block.BlockAir;
 import cn.nukkit.block.BlockLiquid;
+import cn.nukkit.block.BlockWater;
 import cn.nukkit.event.player.PlayerBucketEmptyEvent;
 import cn.nukkit.event.player.PlayerBucketFillEvent;
 import cn.nukkit.level.Level;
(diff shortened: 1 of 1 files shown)


Creeperface01:
whata about for() loop?

Creeperface01:
for(int side : new int[]{Vector3.SIDE_WEST, Vector3.SIDE_SOUTH, Vector3.SIDE_EAST, Vector3.SIDE_NORTH}) {
                        Block b = target.getSide(side);

                        if(b.getId() == STILL_WATER) {
                            level.setBlock(b, new BlockWater());
                        }
                    }

alistairwatts:
That's fine.  Is it worth making the int array private static final on the class to avoid the array being created for each invocation of the method?

alistairwatts:
So, adding to the ItemBucket class:
```
private static final int[] SIDES_NSEW = new int[]{ Block.SIDE_NORTH, Block.SIDE_SOUTH, Block.SIDE_EAST, Block.SIDE_WEST };
```
and then updating the contents of the method to be:
```

for (int side : this.SIDES_NSEW) {
    Block b = target.getSide(side);
    if (b.getId() == STILL_WATER) {
        level.setBlock(b, new BlockWater());
    }
}
```
I'm not really sure what the conventions are for the project.  Are there any particular ones for the project?  I'm just fixing problems as best I can as I encounter them.

Snake1999:
Did the issue solved?

I saw the title `...`

alistairwatts:
The code change solves the issue, but we're discussing the best way to write the code.

alistairwatts:
I've updated the code based on the feedback.

PR #1475, opened 2017-02-10
Update for 1.0.3 release

yescallop:
- Changed Minecraft version string to 1.0.3
- Fixed wrong TransferPacket port data type
- Added player transfer api


Snake1999:
Oh no, you have forgot:
1. You can transfer to both a URN or an IP address. For example, "example.com", 19132
2. Follow-up work should be done after transfer success. e.g., player should be closed (server-side disconnected)

But if you want to make everything in this way for only String and Integer IP/PORT supported without accepting Java API InetSocketAddress, OKAY.

Snake1999:
However, merge first :)

PR #1479, opened 2017-02-10
Default generation. Fix generation of Swamp Biome.

NycuRO:


NycuRO:
https://github.com/Nukkit/Nukkit/pull/1481

PR #1480, opened 2017-02-10
Fix player data saving

Creeperface01:


MrPowerGamerBR:
So this was broken since I implemented it? Wow... I really should've tested that before sending a pull request.

(Okay, I think I forgot to test that because someone requested me to add that to Nukkit, so I just added and didn't check if it was 100% working)

PR #1485, opened 2017-02-14
Fix entity initializing

Creeperface01:
Finally!!! :D i could not figure it out due to ignored exception...


Creeperface01:
or should i add this exception only to debug?

MagicDroidX:
Well, it should be fine

PR #1501, opened 2017-02-27
Transfer Event and transfer command!

AnkitM252:
Transfer event and Transfer command (/transferserver)


AnkitM252:
I'll reopen after all error fixes!

PR #1502, opened 2017-02-27
Transfer event and command!

AnkitM252:


src/main/java/cn/nukkit/Player.java
```diff
@@ -4567,15 +4566,19 @@ public void setSprinting(boolean value, boolean setDefault) {
...
-    public void transfer(InetSocketAddress address) {
-        String hostName = address.getHostName();
-        int port = address.getPort();
-        TransferPacket pk = new TransferPacket();
-        pk.address = hostName;
-        pk.port = port;
-        this.dataPacket(pk);
-        String message = "Transferred to " + address + ":" + port;
-        this.close(message, message, false);
+    public boolean transfer(String address, int port) {
```
Snake1999: Are you really sure to EDIT THE APPLICATION INTERFACE as what you would like to be?

AnkitM252: I'll add it back!

PR #1514, opened 2017-03-07
Item Frame error

imjack:
The packet received doesn't seem to send the slot which results in an error. This fixes the error.


Snake1999:
Testing

PR #1524, opened 2017-03-10
Temporary fix for player movement

### LINKED ISSUE #1519: Other players are not working
### Issue description
It seems that the player you are manipulating does not seem to be moving from other people, and it seems that you are not moving from others.

I do not know because I have not logged in to anything other than my server, but it works normally with Wi-Fi multi.

1. join server

### OS and versions

* Nukkit Version: #1517 

* Client Edition: Android

* Client Version: 1.0.4

Tee7even:
Fixes #1519

### DIFF
--- a/src/main/java/cn/nukkit/level/Level.java
+++ b/src/main/java/cn/nukkit/level/Level.java
@@ -840,7 +840,19 @@ public class Level implements ChunkManager, Metadatable {
             int chunkX = getHashX(index);
             int chunkZ = getHashZ(index);
             for (MovePlayerPacket packet : this.playerMoveToSend.get(index).values()) {
-                this.addChunkPacket(chunkX, chunkZ, packet);
+                //this.addChunkPacket(chunkX, chunkZ, packet);
+
+                //temp fix
+                //todo: find a proper solution
+                List<Player> players = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


AndreySK4:
The fix doesn't work on my server :(

Creeperface01:
@Tee7even what about direct move sending instead of current one?

Tee7even:
@Creeperface01 do you mean with `player.dataPacket` or what? I don't know exactly, but there's some stuff related to packet batches in `broadcastPacket`, that may be important.

Creeperface01:
No i mean from entity class addMovement() method

Tee7even:
I'll check it.

Tee7even:
@Creeperface01 yeah, it works. That leaves a quesion of what to do with the methods utilizing MovePlayerPacket.

imjack:
So what is there a fix going to be pushed to master?

Creeperface01:
probably direct packet sending :D

Creeperface01:
we can use getVierwers() instead of chunk players i think

Tee7even:
>we can use getVierwers() instead of chunk players i think

There's no sense in changing anything than removing an overrided method for moving in `Player`.

PR #1526, opened 2017-03-10
' ' Fix error in many plugins

kvetinac97:
As of 1.0.4, ' ' char now isn't blocked in MC: PE. However, when used with Config or in some plugins, it causes many errors

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -1983,7 +1983,7 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
                         if ((c >= 'a' && c <= 'z') ||
                                 (c >= 'A' && c <= 'Z') ||
                                 (c >= '0' && c <= '9') ||
-                                c == '_' || c == ' '
+                                c == '_'
                                 ) {
                             continue;
                         }


Tee7even:
Xbox gamertags can contain spaces. And what "error" exactly are you talking about?

MrPowerGamerBR:
@Tee7even probably because YAML files that stores like this "data.playername.***" will break if someone logs in with a " " name.

Pub4Game:
In my plugins, everything works perfect

image

Creeperface01:
you can make name filter from plugin

kvetinac97:
Yep, fixed it like that

Tee7even:
@Pub4Game are those passwords?

Pub4Game:
@Tee7even Yes, it's passwords. 😅

MrPowerGamerBR:
@Pub4Game ew, you store passwords in plain text, that's a huge security risk.

Pub4Game:
@MrPowerGamerBR It's probably up to me to decide how and what to do, okay?

MrPowerGamerBR:
@Pub4Game did I complain? No, I just said that it is a huge security risk, after all it is your server so I don't care if you store every password in plain text, just remember me to never play on your server.

Creeperface01:
In the past I stored password also as plain text simply because i used even pocketmine and pocketmine hashes are not compatibile with java hashes and i didnt have a web server and enough experience to do that via external server :D

Creeperface01:
@MrPowerGamerBR im using special password for all those servers :D because some of them dont hash passwords :D

Issue #1520, opened 2017-03-10
No ItemBlazeRod.java


Leonidius20:
There is no ItemBlazeRod.java in the lastest (560) nukkit build, so if i want to get Blaze Rod's name (getName()), I get "Unknown". Maybe this file need to be just like, for example, ItemLeather.java, but with other ID and name. Please, add this file. Thanks

kazuemon:
Is this problem only solved by adding ItemBlazeRod.java which copied ItemEgg.java and rewritten?

kazuemon:
https://github.com/kazuemon/Nukkit/commit/5be073578649d34e9feb456834be1a473b9b3e17
https://github.com/kazuemon/Nukkit/commit/2d3c02a0f27e4f1bc5df77d2567bfb6de03bf5ec

like this

Leonidius20:
@kauzemon yes

PR #1529, opened 2017-03-13
add BlazeRod.java

kazuemon:
https://github.com/Nukkit/Nukkit/issues/1520

Since I am not used to using Github, it may contain other than the necessary files.


kazuemon:
miss

PR #1537, opened 2017-03-16
Give purpose to Stringenum

kvetinac97:
Now, the stringenum option HAS sense.
purpose


Pub4Game:
What is it?

kvetinac97:
E.g. in command /time set, there are 2 overloads:

1. /time set <int>
2. /time set <time: string>

When using stringenum, you can define which strings MUST be used
so: new CommandParameterEnum("time", new String[]{"day", "night"});
will result in MCPE client forced to use /time set day or /time set night (otherwise Invalid command syntax will be thrown)

kvetinac97:
I will rework it & make new request

PR #1539, opened 2017-03-17
Make using commands easier for players

kvetinac97:
This commit introduces two new fields to ArgType called stringenum : enum_values (you can predefine which strings can be used by player) and enum_type (blockList, itemList... so MCPE will automatically display list of all supported blocks, items)

Screenshots:
purpose2
purpose4
purpose5
purpose6

Example of usage in plugin:

purpose
purpose3

### DIFF
--- a/src/main/java/cn/nukkit/command/defaults/GiveCommand.java
+++ b/src/main/java/cn/nukkit/command/defaults/GiveCommand.java
@@ -24,6 +24,18 @@ public class GiveCommand extends VanillaCommand {
                 new CommandParameter("meta", CommandParameter.ARG_TYPE_INT, true),
                 new CommandParameter("tags...", CommandParameter.ARG_TYPE_RAW_TEXT, true)
         });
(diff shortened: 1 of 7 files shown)


src/main/java/cn/nukkit/command/defaults/GiveCommand.java
```diff
@@ -17,22 +17,11 @@ public GiveCommand(String name) {
         super(name, "%nukkit.command.give.description", "%nukkit.command.give.usage");
         this.setPermission("nukkit.command.give");
         this.commandParameters.clear();
-        this.commandParameters.put("toPlayerById", new CommandParameter[]{
+        this.commandParameters.put("default", new CommandParameter[]{
```
Tee7even: Could you please return deleted options, so an item can be given both by id and by name?

kvetinac97: I was thinking about it, but wasn't on computer at time, I'll return them

PR #1557, opened 2017-03-21
fix crafting snow block in inventory

### LINKED ISSUE #1533: No snow block crafting recipe  in inventory
There is no recipe for making snow block in the inventory. I can only make it on the crafting table. But in vanilla single player game there are both recipes.
screenshot_2017-03-14-15-34-29

Leonidius20:
Now you can craft snow blocks in inventory without crafting table. Fixes #1533


Leonidius20:
My old commits are here too, i don't know why. The changes are in the CraftingManager.java

Tee7even:
>My old commits are here too, i don't know why. The changes are in the CraftingManager.java

Because of the squash merge, the best you could do is to create new branches for each contribution, propose a pull request from them and then, when the pull request is merged, delete the branch and update your repo from upstream.

PR #1567, opened 2017-03-26
Improvements for soups

Leonidius20:
1. Now you'll eat beetroot soup _without_ the bowl.
2. After eating the soup bowl will be placed to slot where the soup was.

### DIFF
--- a/src/main/java/cn/nukkit/item/food/FoodInBowl.java
+++ b/src/main/java/cn/nukkit/item/food/FoodInBowl.java
@@ -17,7 +17,8 @@ public class FoodInBowl extends Food {
     @Override
     protected boolean onEatenBy(Player player) {
         super.onEatenBy(player);
-        player.getInventory().addItem(new ItemBowl());
+        int slot = player.getInventory().getHeldItemSlot();
+        player.getInventory().setItem(slot, new ItemBowl());
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


Tee7even:
Ummm...
image
image
image

kazuemon:
Almost the same, how about using Item.get ()

Tee7even:
I guess, that's not the problem, more likely, that the problem is that `setItem` acts not like @Leonidius20 expected.

Leonidius20:
The problem is that getHeldItemSlot() returns -1 if the slot is empty and I can't understand why.

xISRAPILx:
@Tee7even Дай борщ!!!

PeratX:
Tests needed.

src/main/java/cn/nukkit/item/food/FoodInBowl.java
```diff
@@ -17,7 +17,8 @@ public FoodInBowl(int restoreFood, float restoreSaturation) {
     @Override
     protected boolean onEatenBy(Player player) {
         super.onEatenBy(player);
-        player.getInventory().addItem(new ItemBowl());
+        int slot = player.getInventory().getHeldItemSlot();
+        player.getInventory().setItem(slot, new ItemBowl());
```
NycuRO: Why setItem. Just use addItem.

PR #1707, opened 2017-04-27
Breaking your plugins. Again.

Tee7even:
This PR introduces some changes that most likely will break your plugins. So you may want to update them.
All the changes of this pull request:
- Integer side constants in `Vector3` are replaced with the `BlockFace` enum.
- Reworked `EntityDamageEvent`:
  - Integer constants are replaced with enums: `DamageCause` and `DamageModifier`.
  - `attack` method of `Entity` class now returns a boolean value: `true` if damage was dealt to an entity and `false` if not.
  - The event is not fired if server determines that no damage was dealt to an entity (for example if player is in creative mode or the last damage was not enough time ago), so the vanilla behaviour now can't be broken by de-cancelling it.
- Redstone by @Creeperface01. There's done pretty much, but not everything works just yet.


Creeperface01:
One question, I think now if you make repeating redstone circuit and chunks gets unloaded, circuit will be stopped. And I don't know how should I fix it.

Tee7even:
Do you mean it will stop forever? Or it will continue when it gets loaded? In the first scenario there's nothing to be fixed - it's okay, servers always work like this. In the second one - well, half of the system has to be reworked probably.

Creeperface01:
repeaters etc are scheduled to check redstone, and update around redstone, if chunk gets unloaded, cheduled update will be cancelled so it will be stopped forever until somthing updates it again

Tee7even:
Did you port redstone from somewhere or not?

Creeperface01:
Emm a bit :D I can look at cuberite how they solved that

Creeperface01:
I have an idea how to solve that

Creeperface01:
We can save some list of blocks that should be updated when chunk will be ticked first time

Snake1999:
We should have a discuss on this PR

Tee7even:
@Snake1999 we can't do that for example with `EntityDamageEvent` - we either store damage cause and modifiers as integer constants or as enums. And I think we should keep enums. Maybe change as much as we can with one shot. 
@Creeperface01 doesn't sound too good for me, but prepare an example, we can discuss that.

Creeperface01:
Redstone should be  ticked only in a certain radius around players shouldnt it?

Tee7even:
In loaded chunks probably, I actually don't know.

vzlomka76:
@Creeperface01 you can do the pistons working? i think they are already ready by 70%

Creeperface01:
@vzlomka76 yes they are, but blockentity is somehow wrong

vzlomka76:
@Creeperface01 you can try to do without BlockEntity?

Creeperface01:
But then players see the puston head invisible

vzlomka76:
@Creeperface01 Oh, Ok ;)

Snake1999:
Ok Lets keep enumerations

Tee7even:
Well, @Creeperface01? Any idea how to fix it or something?

Creeperface01:
Pistons? It seems that MCPE sends exactly same data so idk whats wrong :/

Tee7even:
No, I am talking about your "circuitry stops forever" stuff.

Creeperface01:
Yes as I said maybe update all repeaters and comparators in the first chunk tick.

Tee7even:
After load?

Creeperface01:
Yes, I think chunks are ticked only in certain radius around players, so redstone (diode) scheduled updates, should be too, but they are maybe

Tee7even:
Seems like a good plan.

Snake1999:
Let's accept the basic functions, and then finish it. After a few months' delay, more and more conflicts will appear.
In a word, we should accept this PR if necessary, ASAP.

Tee7even:
Yeah, I also thought of merging it, we didn't have working redstone anyway. But it is a bit bad to merge an unfinished (bugged) feature.

wbzhang2594:
Strongly suggest to increase the Nukkit version while making so huge change. Rather than keeps this version and breaks other plugins' building.

Creeperface01:
@Tee7even or we can update all diodes using blockentity, but in vanilla its solved differently

PR #1737, opened 2017-05-27
Fix method params conflict in BaseLang

yescallop:
Fixed weird message in Achievement and ResourcePackManager due to this conflict.


Snake1999:
Maybe we can change the old method name instead of all the calling ways.
Remember to @Deprecated

PR #1742, opened 2017-05-31
Compass position is updating now

kvetinac97:
There were 2 uknown fields in SetSpawnPositionPacket. One of them, the VarInt (not UnsignedVarInt) one meant spawn type (player or world).
Now, when player switches world, or /setworldspawn command gets called, position compass is pointing to gets updated also (so the compass serves purpose now).


Snake1999:
Great

Issue #1743, opened 2017-06-02
Update Nukkit to support client verstion 1.1.0


dcmwai:
Unable to join server when client version is 1.10 (Win 10 PE/iOS

---
Edit by @Snake1999:

**We need your help for testing our software for client 1.1.0!** Compile code on https://github.com/yescallop/Nukkit/tree/1.1.0 and try it out~ After tested OK, our main branch will be updated to 1.1.0. 

We are now working 1.1.0 on: https://github.com/yescallop/Nukkit/tree/1.1.0
Any contributions please go to the site above. We welcome every bit of your work.

Go on nukkiting out our code :)

---

As our new branch is bleeding - still under construction and maybe buggy or unfit for production use, no jar file are available for 1.1.0 yet unless you do the compile work by yourself. @PetteriM1 

---

yungtechboy1:
👍

imjack:
That's because it's not updated to the latest version

MasterlazorX:
It's not updated to to 1.1 yet.

kvetinac97:
Anyways, somebody should update it. @yescallop ?

MasterlazorX:
I'm sure there's a BUNCH of protocol changes to mcpe 1.1.

Creeperface01:
or @kvetinac97 ?

jrodr10:
We're gonna need a binary rewire & nbt rewire & protocol change to 113 ._.

DaMatrix:
...and this is where nukkit dies. with no active devs and a massive update like this, we ain't gonna be getting an update at this rate. :/

RednedEpic:
1.1 came out YESTERDAY. I bet the developers are hard at work at trying their best to update Nukkit to 1.1. As a developer, once there is a new update of Minecraft to update plugins to. It definitely takes a lot longer than a day to update. And if you don't believe me, go ahead, Fork Nukkit and update it yourself.

DaMatrix:
i know more than a bit about what's required, i'm making a general commentary that the original devs are pretty much dead and nukkit doesn't get much any more other than occastional version bumps and a PR every once in a while.

boybook:
I'm trying to write about it but lots of problems.

ScraM-Team:
Let us know how we can help with 1.1.0.  We're big Nukkit fans here on the ScraM team.

Creeperface01:
@yescallop is quite experienced so maybe he could help us with update

[...]

PR #1754, opened 2017-06-14
It's time to update to 1.1.0

Snake1999:
Still buggy on:

- [x] Client crash when joining in (Confirmed as client issue)
 - [x] Block won't break

After fixed, our 1.1 branch is nore stable and at least, able to play. 
Thank you for contributing to Nukkit :)
https://github.com/Nukkit/Nukkit/issues/1743


jrodr10:
Still one more bug to fix.. Players whom are OP cannot join

AppleDevelops:
@jrodr10 I can join while I’m opped, try without plugins

Goodstuff4UonYT:
@AppleDevelops Did you use Win10? 

----
And please don't forget to implement the function which "fixes"  the chunk bug :D

AppleDevelops:
@Goodstuff4UonYT I used W10

jrodr10:
Same result @AppleDevelops No Plugins running the latest Nukkit Version , Also @Goodstuff4UonYT try having your worlds spawns located not on the same cords in both worlds it will fix the chunk loading glitch

sero583:
Btw bed colors turn automatically after placing to white just wanted to say

Snake1999:
@sero583 try express your problem with your native language as I couldn't understand what "Btw bed colors turn automatically after placing to white just wanted to say" means

sero583:
@Snake1999 What I try to say is, in MCPE now dyeable beds are avaible as you already probably know and if you for example place a blue bed on the ground the bed block directly turns into a white bed

AppleDevelops:
It’s because beds are now tiles

sero583:
@AppleDevelops so you probably need to set NBT values on entity and this explains why its white I think because it's the default NBT of it or something like that, or?

xpyctum:
Nice work

Snake1999:
We are now making game available to join. Detailed modifications such as "dye beds" will come after that

jrodr10:
My bedwars server had player tp to spawn then back to the arena on death to fix the chunks not loading 😂

jrodr10:
OP Players Crash on Join still @snake1999

sero583:
@jrordr10 for me it works fine, why dont you try maybe 1. disabling plugins 2. Switching to another map 3. Try to join with non op to ensure that it really crashes only on OP?

larryTheCoder:
@Snake1999 I think you aware that Boats are broken. Player need to ride on the boat not under the boat. I found this bug when trying to ride the boat. I start figure out that this is SetEntityLink error... or client bug?

vzlomka76:
@Snake1999 I don't have problem with op-join. I have op and just join. But sometimes (all players) have crash client on join. Usually client crash when disconnect from server with all reasons

kpwnh:
I can join with OP in my server. The only problem is the client doesn't get any commands, for example, even /help doesn't work.

vzlomka76:
@kpwnh i don't have this problem, do you have any plugins?

jrodr10:
@sero583 nope just op players can connect from the version im using https://github.com/jrodr10/Nukkit

kpwnh:
It seems the problem is with Google Cloud servers. Has anybody tested Nukkit in Google Cloud?

Local Nukkit works properly for me

jrodr10:
I use OVH Dedicated Servers I've got two

Specs:
Intel  i7-7700K OC
4c/8t - 4.7 GHz/5.00 GHz
64GB DDR4 2400 MHz
SoftRaid  2x450GB  NVMe
250 Mbps  bandwidth

larryTheCoder:
Can we change the API version? There a lot of changes were made since v1.0.4

Snake1999:
Now let's make this change

Issue #1789, opened 2017-07-10
Nukkit is using many resources


cloud9350:
### Issue Description

Write a short description about the issue
Nukkit has using Heavy resources

### Steps to Reproduce the Issue

1. Start Nukkit (FLAT map)
2. Wait 8hours (Can players join)
3. [![1499729173799.png](https://s23.postimg.org/czv2l2l4r/1499729173799.png)](https://postimg.org/image/574et3f5j/)

### OS and Versions

* Nukkit Version: Last commit 959af73 

* Java Version: Java 8 Latest. I'm banned hosting (Many using resources)
```
??
```

* Host Configuration:

| Item | Value |
|:----:|:-----:|
| Host OS | Windows |  
| Memory(RAM) | 64 GB | 
| Storage Size | 1TB | 
| Storage Type | SSD | 
| CPU Type | Intel Xeon X5650 | 
| CPU Core Count | 24 cores 47 threads | (Can't control threads)
| Upstream Bandwidth | 1Gbps | 

* Client Configuration: 

| Item | Value |
|:----:|:-----:|
| Client Edition | Android | 
| Client Version | 1.1.3 | 

### Crashdump, Backtrace or Other Files

No

cloud9350:
No plugin

ScraM-Team:
We haven't experienced this on our servers.  Have you set a max heap size?  "-Xmx2g" on the java command line will limit the memory used to 2GB heap + a little more for the VM, etc. (roughly 2.2GB total).

We are not using the latest commit, however.

cloud9350:
@ScraM-Team [![20170711_104613.jpg](https://s3.postimg.org/edmios9sj/20170711_104613.jpg)](https://postimg.org/image/wgflg05n3/)
Not working (Home computer)

cloud9350:
Oops wrong enter command lol
<a href='https://postimg.org/image/xfqneyu0l/' target='_blank'><img src='https://s10.postimg.org/dxvzz0x2x/14997377687681934237506.jpg' border='0' alt='14997377687681934237506'/>

ScraM-Team:
-Xmx2g must come before the -jar parameter.  Try this:

> java -Xmx2g -jar nukkit.jar

2g is for 2 GB max heap, you can adjust to suit your needs (m works for MB too)

cloud9350:
@ScraM-Team It's working

Snake1999:
Close?

cloud9350:
[...]

PR #1790, opened 2017-07-11
Reduce memory usage, phase 2 (Zlib)

Snake1999:
We have reduced memory usage by enhanced VarInt in parse 1 (https://github.com/Nukkit/Nukkit/pull/1776).
Zlib is eating more and more memory as well. That is caused by creating too much instances by the deflater and inflater. Doing enhancements :)
Also see https://github.com/Nukkit/Nukkit/issues/1789.

Tests have done for commit on "Reduce byte array creation". In each test round, one player joins the server, stay for 30 seconds, then run to the horizon for 30 seconds, then quit the game. *Controlled group using software before this PR, experimental group using software after "Small improvement for Zlib"*. After 5 test round each group, *450 MiB (max VM submit) and 310 MiB (max VM consume) of memory is used by controlled group*, while *300 MiB (max VM submit) and 240 MiB (max VM consume) is for experimental group*. **About one quarter of memory is saved**. 

Further enhancement will be applied using Netty-provided ByteBuf.

- [x] Reduce byte array creation
- [x] Test class for Zlib
- [ ] Replace byte array and its streams to ByteBuf provided by Netty

### DIFF
--- a/src/main/java/cn/nukkit/utils/Zlib.java
+++ b/src/main/java/cn/nukkit/utils/Zlib.java
@@ -4,25 +4,12 @@ import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.Map;
-import java.util.Optional;
-import java.util.TreeMap;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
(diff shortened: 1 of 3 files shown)


Snake1999:
ByteBuf provided by Netty will be pushed in another PR.

Issue #518, opened 2016-02-20
Bug with creative inventory


Pub4Game:
If you open the inventory in creative, he is filled with blocks, but should be empty

![image](https://cloud.githubusercontent.com/assets/11244874/13199752/11437cd0-d83f-11e5-8baf-47c18408fcba.png)

PR #1808, opened 2017-07-18
Send Creative Inventory on Player First Spawn

### LINKED ISSUE #1786: In creative mode, sometimes no items in creative inventory.
### Issue Description
There is sometimes nothing inside the creative inventory! I use default gamemode as creative.

### Steps to Reproduce the Issue

1. Join the server with creative mode
2. Open the inventory
3. Nothing inside

### OS and Versions

* Nukkit Version: https://github.com/Nukkit/Nukkit/commit/58628743b5254be17f28bc247eef47ba87f20f98 

* Java Version: 
```java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.131-b11, mixed mode)
```

* Host Configuration: 

```root@localhost>uname -a
Linux cloud 2.6.32-642.el6.x86_64 #1 SMP Tue May 10 17:27:01 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```
| Item | Value |
|:----:|:-----:|
| Host OS | Linux |  
| Memory(RAM) | 16 GB | 
| Storage Size | 200 GB | 
| Storage Type | HDD | 
| CPU Type | ?? | 
| CPU Core Count | ?? cores ?? threads | 
| Upstream Bandwidth | ?? | 
Collecting, TODO finish

* Client Configuration: 

| Item | Value |
|:----:|:-----:|
| Client Edition | iOS, Android and Windows 10 | 
| Client Version | 1.1.2 | 

### Crashdump, Backtrace or Other Files

```
<!--- Copy and paste your backtrace here... -->
```

Christian7573:
I got bored, so I attempted to fix #1786

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -759,6 +759,15 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
         this.sendData(this);
         this.inventory.sendContents(this);
         this.inventory.sendArmorContents(this);
+        
+        //Send creative inventory
+        if (this.getGamemode() == 1) {
+            ContainerSetContentPacket containerSetContentPacket = new ContainerSetContentPacket();
+            containerSetContentPacket.windowid = ContainerSetContentPacket.SPECIAL_CREATIVE;
+            containerSetContentPacket.eid = this.id;
+            containerSetContentPacket.slots = Item.getCreativeItems().stream().toArray(Item[]::new);
+            this.dataPacket(containerSetContentPacket);
+        }
 
         SetTimePacket setTimePacket = new SetTimePacket();
         setTimePacket.time = this.level.getTime();
--- a/src/main/java/cn/nukkit/event/player/PlayerEvent.java
+++ b/src/main/java/cn/nukkit/event/player/PlayerEvent.java
@@ -8,7 +8,7 @@ import cn.nukkit.event.Event;
  * Nukkit Project
  */
 public abstract class PlayerEvent extends Event {
-    public Player player;
+    protected Player player;
 
     public Player getPlayer() {
         return player;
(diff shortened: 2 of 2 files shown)


jonathanpeppers:
Did this used to work fine? Did perhaps the PR here break it: https://github.com/Nukkit/Nukkit/issues/518

I haven't used nukkit for a while, and updated to see this issue happening. @sero583 any suggestions on what to do-- I can take a stab at it.

sero583:
@jonathanpeppers A research would be good to know when it happens and taking a look at the current inventory sending code, for me it used to happen sometimes but it wasn't really a major issue thats why I didn't reported it anytime. It happened for me sometimes also after gamemode changing.

sero583:
Oh yes that change which made the Player variable from protected to public is not good, because plugins can actually edit the variable how they like to.

jonathanpeppers:
@sero583 

Here is how it has been happening for me (this is issue #1786):
- Wipe the `players` directory
- Run Nukkit with `defaultgamemode` set to `creative`
- First player logs in, has inventory as usual
- Any subsequent player joining has no inventory--this includes if the first players leaves and rejoins

I was finding I could only get the inventory back if I deletes all the player `.dat` files, and then only the first player could get them.

I have Minecraft: PE 1.1.4 on iOS.

sero583:
Also Event changes made which shouldn't be made.

Christian7573:
Ok, I'll do some research. Also, if you're referring to the weird public / protected change in PlayerEvent, that was something from when I first downloaded nukkit and I undid it in the next commit.

augustoeliezer:
Thanks @Christian7573 

See https://github.com/Nukkit/Nukkit/pull/1842, same fix but add for each user in server.

Issue #1508, opened 2017-03-03
Potions do not work on entity collision/hit


sero583:
### Issue description

Potions doesn't apply on anything any effect (throwables).

### Steps to reproduce the issue

1. Test a throwable potion
2. See
3. Notice that nothing happens.

### OS and versions

Nukkit Version: This server is running Nukkit version 1.0dev "??(Apple)?(Pie)" (API 1.0.0)/works probably on any version
Java Version: java -version
openjdk version "1.8.0_121"
OpenJDK Runtime Environment (build 1.8.0_121-8u121-b13-0ubuntu1.16.04.2-b13)
OpenJDK 64-Bit Server VM (build 25.121-b13, mixed mode)
Host OS: Ubuntu LTS 16.04.1
Doesnt matter below

Host Configuration: ?? GB RAM, ?? GB ??(SSD/HDD/...) drive, ?? core CPU from ??(Intel/AMD/...), ?? Mb internet upload
Client Edition: ??
Client Version: 1.0 beta (doesnt matter probably though)

JackboyPlay:
Drop it into the air and stand under the potion.

sero583:
I found more out about that, they work particually; means if they are dropped into ground and in the effect radius is a Entity it receives the effect, but if you hit a Entity with the Potion it's disappearing.

Snake1999:
Reproduced

sero583:
Whoops came on that button, sorry.
@Snake1999 I looked in the source codes and have no idea what causes it, good luck at finding it.

Creeperface01:
I can fix it

sero583:
@Creeperface01 Please try to and open a PR then

sero583:
^ bumppp

Tee7even:
>^ bumppp

It won't make this any faster. Do not bump the issues.

sero583:
Okay.

sero583:
@Tee7even Do you think this has something to do with the ``attack(EntityDamageEvent source);`` to do (its not overridden in EntityPotion) or something with collision? I think it's ``attack(EntitDamageEvent source);`` because there it should splash

sero583:
So I was looking where the reason lies and the only way I could think of is it lies that the potion gets closed as soon as it performs the super.onUpdate(int); call, because then the closed boolean is set to false and yes the onUpdate is cancelled and cannot perform. This would explain to me why it does 0 damage because the attack(); method is called. @Creeperface01 @Snake1999

PR #1811, opened 2017-07-23
Add the support to change Apply Effect in PotionApplyEvent

boybook:
Add API to get and change the apply effect when a potion applied on an entity.


sero583:
Sorry if I distrub, but that issue will be this issue fixed anytime soon? (#1508)

Issue #1795, opened 2017-07-11
Damage by Projectile items


asyven:
### Issue Description

All of items who enxtends EntityProjectile give damage for player or entity.

### Steps to Reproduce the Issue

1. Get XPbottle, Snowball or Enderperl, etc
2. Direct on player or entity and use item

ScraM-Team:
Is there a complete list (on the Minecraft wiki, perhaps), of which projectiles shouldn't cause damage?  When making this change it would be easier to have the complete list so it can be fixed all at once.

asyven:
@ScraM-Team
only bow must cause damage
but some items set knockback

sero583:
Potion also don't makes effect afaik, it just makes a hit

Angelic47:
Fixed by @ScraM-Team

PR #1820, opened 2017-07-27
Fix for Issue #1795: Non-Arrow projectiles shouldn't cause damage

ScraM-Team:
This change fixes Issue #1795 

It changes the damage logic to allow base damage to be overridden on a per-projectile basis, defaulting to zero (no damage).  Only arrow overrides this, so only arrows cause damage.

I reused the old projectile damage amount of 2, now just for arrows.

I have tested by shooting arrows and snowballs straight up into the air in survival mode.  Arrows still cause damage when they fall, but snowballs no longer do.

Let me know how it looks, this is my first Nukkit commit so I could easily have missed something.

### DIFF
--- a/src/main/java/cn/nukkit/entity/projectile/EntityArrow.java
+++ b/src/main/java/cn/nukkit/entity/projectile/EntityArrow.java
@@ -91,7 +91,8 @@ public class EntityArrow extends EntityProjectile {
         return base;
     }
 
-    private double getBaseDamage() {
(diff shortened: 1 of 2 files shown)


Angelic47:
Nice job

ScraM-Team:
Woohoo, my first Nukkit PR accepted!  :-)  Thanks @Angelic47 and @kvetinac97

Issue #1817, opened 2017-07-25
The console can not enter a command and the main thread stuck


WetABQ:
### Issue Description
Main thread deadlock
Server console stuck, can not enter the command, the player will not be quit but in addition to asynchronous thread plugin can operate other  core of the main functions are not working
Even if a completely new server will exist(This server is completely new without plugins and settings)
服务器控制台卡死，无法输入命令，玩家不会掉线但除了异步线程的插件可以运作其他都不行并且核心的主要功能都无法运作
就算是新开一个服务器这个问题还会存在(这个服务器完全是新的不存在插件和设置)
And no player presence can also cause this problem
并且没有玩家存在也可能造成这个问题

### Steps to Reproduce the Issue
1. Start the server
2. Wait....
3. Main thread deadlock.
But this is a probabilistic event.
And not all servers will happen.
But once the encounter will continue to trigger.
### OS and Versions
* Nukkit Version: d028383d4920468f0e15fbbc44531e2209855464
* Java Version: openjdk version:
```
openjdk version "1.8.0_131"
OpenJDK Runtime Environment (build 1.8.0_131-8u131-b11-0ubuntu1.16.04.2-b11)
OpenJDK 64-Bit Server VM (build 25.131-b11, mixed mode)
```
* Host Configuration: 

| Item | Value |
|:----:|:-----:|
| Host OS | Ubuntu |  
| Memory(RAM) | 4 GB | 
| Storage Size | 128 GB | 
| Storage Type | SSHD | 
| CPU Type | E5-2695v3  | 
| CPU Core Count | 4 cores 4 threads | 
| Upstream Bandwidth | 1Gbps | 

* Client Configuration: 

| Item | Value |
|:----:|:-----:|
| Client Edition | Android&IOS&Windows10 | 
| Client Version | 1.1+(1.1.3) | 

### Crashdump, Backtrace or Other Files
No error, you do not know when the thread stuck
完全无报错,你根本不知道什么时候线程卡死了

ScraM-Team:
I am seeing the same thing.  This thread seems to be in common between each of my stack traces:

`"main" #1 prio=5 os_prio=0 tid=0x00007f7154009000 nid=0x4a0 runnable [0x00007f715b6ea000]
   java.lang.Thread.State: RUNNABLE
        at java.lang.ThreadLocal$ThreadLocalMap.getEntryAfterMiss(ThreadLocal.java:444)
[...]

PR #1826, opened 2017-07-30
Fix for issue #1817 -- main thread stuck (rollback Zlib)

ScraM-Team:
This rolls back a change to Zlib.java which caused infrequent lockups of the main thread.  While infrequent, on a public server it was pretty serious when it happened since commands can't be entered on the console and no new users can log in.

Since the old code being restored created new object frequently, expect more frequent garbage collection if you have been using builds from the past 2-3 weeks.

### DIFF
--- a/src/main/java/cn/nukkit/utils/Zlib.java
+++ b/src/main/java/cn/nukkit/utils/Zlib.java
@@ -9,24 +9,47 @@ import java.util.zip.InflaterInputStream;
 
 
 public abstract class Zlib {
-    
+
     public static byte[] deflate(byte[] data) throws Exception {
         return deflate(data, Deflater.DEFAULT_COMPRESSION);
     }
 
     public static byte[] deflate(byte[] data, int level) throws Exception {
-        Deflater deflater = getDef(level);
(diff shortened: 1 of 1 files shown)


Snake1999:
Ah! You guys destroyed my work! Now we went back to the days when GBs of memory eaten up in seconds!
Why shared buffers be cancelled? I cannot reproduce your problem!

Snake1999:
You are avoiding the real problem just by doing more and more rollbacks - rollbacks of history. 
Have you ever considered why should a project targeted to enhance performance chose to use "tick"? Just go modify the tick system and develop a executor service to take place of it

ScraM-Team:
@Snake1999 , I agree with you, this change increases memory use.  I want to help put back your change.  

Step 1: Put back your change with a config file setting to allow it to be turned on or off.  This way people like you who never had server freezes could use your memory-saving fix.

Step 2: Try an alternate version of your fix, perhaps using a shared pool of buffers instead of ThreadLocal.  Conceptually it is the same as your fix, but maybe it would avoid the freezes that I and others were seeing.

(By the way, I never saw the freezes either in a non-production environment.  It only happened in production, with lots of users logging in.  I agree it is very hard to debug.)

PR #1840, opened 2017-08-03
[Packets] Update Packets and More Packets!

NycuRO:
I'm observed to Issues and 90% is datorated by Packets Incomplete.
And for that, i'm trying to port PM-MP Packets (It's so good, completly and fully functional).
Now i'm trying start Port AddEntityPacket. 

Anyone can help me 👍 ( @Creeperface01 You need help me! 😆 ).

Note: I'M USED INTEL NOW.

**Changes**
1. I'm deleted this.reset() method because is more redundant and is not good. (Dylan says that).
2. I'm changed encode() and decode() to encodePayload() and decodePayload().

### DIFF
--- a/src/main/java/cn/nukkit/network/protocol/AddEntityPacket.java
+++ b/src/main/java/cn/nukkit/network/protocol/AddEntityPacket.java
@@ -32,13 +32,12 @@ public class AddEntityPacket extends DataPacket {
     public final Object[][] links = new Object[0][3];
 
     @Override
-    public void decode() {
+    public void decodePayload() {
 
     }
 
     @Override
-    public void encode() {
-        this.reset();
+    public void encodePayload() {
         this.putVarLong(this.entityUniqueId);
         this.putVarLong(this.entityRuntimeId);
         this.putUnsignedVarInt(this.type);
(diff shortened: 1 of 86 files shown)


NycuRO:
Help, please? -_-

sero583:
@NycuRO Your changes are non-sense ofcourse they won't work. Please inform yourself better about what ``abstract`` means and IS in java. Also changing function names should be avoid better, because it breaks plugins which are using that part of the API.

NycuRO:
@sero583 Now i'm ported from PM-MP. I know this is not definited as abstract function, but in php works...

sero583:
@NycuRO Porting something from PHP to Java shouldn't be done, better re-write it yourself also even if it works in PHP this is Java.

NycuRO:
@sero583 I know that, but nukkit is similar as pmmp.

sero583:
@NycuRO You should stop doing non-sense stuff since @yescallop said that everyone should work at his repo. I would suggest closing this non-sense stuff you change variable types how you like to and expect that it will work?

sero583:
@NycuRO ``https://github.com/yescallop/Nukkit/tree/1.2`` here everyone who likes to contribute on the update should work, informing yourself before doing some weird things is better than rushing into work.

NycuRO:
@sero583 Ok..

PR #1842, opened 2017-08-04
Fix for #1786 - creative inventory not generating

MitchellBot:
- Added sendCreativeContents function to PlayerInventory
- Removed ContainerSetContentPacket logic from ProcessLogin()
- Added call to sendCreativeContents to player during doFirstSpawn()
- Added call to sendCreativeContents to the reload command to regenerate all players' creative inventories

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -754,6 +754,14 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
         this.inventory.sendContents(this);
         this.inventory.sendArmorContents(this);
 
+        /*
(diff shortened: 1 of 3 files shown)


NycuRO:
@MitchellBot Same bug it's and other Gamemodes, but it fixed by Client and Packet. Can you do and for other Gamemodes for remediate "minore bugs"

NycuRO:
@MitchellBot The issue it's and on Survival and Spectator, but this bug it's auto fixed by client, but it's not correct. Need fix and by server.

Add this and for Survival and Spectator Gamemode.

NycuRO:
@MitchellBot I have a hack-fix for that, much easy. 👍

PR #1862, opened 2017-08-14
Fixed PlayerInventory#setArmorContents

Him188:
PlayerInventory#setArmorContents can only change chestplate because of a bug in `for (int i = 0; i < 4; ++i) `:
`this.setItem(this.getSize() + 1, items[i]);`
it should be written as:
`this.setItem(this.getSize() + i, items[i]);`


NycuRO:
What is the change, it's "i"?

Him188:
Right

NycuRO:
Oh looool.

Issue #1861, opened 2017-08-11
Explosions can destroy bedrock


sero583:
### Issue Description

TNT destroys bedrock except the mid it was placed to and placing for example over bedrock lets say over ground where also destroyable blocks are makes the explosion effect less. What I try to say is if TNT stays over an undestroyable block, it does not even checks if in near are some else blocks which should get destoryed. Also item entities which falls into void does not get removed.

### Steps to Reproduce the Issue

1. Generate normal world
2. Place Bedrock or Obsidian somewhere and over it TNT and activate it
3. Go to MCPE's offline mode and try it also there and see the difference
4. Try on y=1 over bedrock an explosion see that it also destroys bedrock.

### OS and Versions

* Nukkit Version: Nukkit 1.0dev ???(Apple)?(Pie)? implementing API version 1.0.0 for Minecraft: PE v1.1.3 (protocol version 113) 

* Java Version: openjdk version "1.8.0_131"
OpenJDK Runtime Environment (build 1.8.0_131-8u131-b11-0ubuntu1.16.04.2-b11)
OpenJDK 64-Bit Server VM (build 25.131-b11, mixed mode)
```
??
```

* Host Configuration: 

| Item | Value |
|:----:|:-----:|
| Host OS | ?? |  
| Memory(RAM) | ?? GB | 
| Storage Size | ?? GB | 
| Storage Type | ?? | 
| CPU Type | ?? | 
| CPU Core Count | ?? cores ?? threads | 
| Upstream Bandwidth | ?? | 

* Client Configuration: 

| Item | Value |
|:----:|:-----:|
| Client Edition | Android | 
| Client Version | 1.1.3 |

NycuRO:
Hmm, wait.

NycuRO:
![image](https://user-images.githubusercontent.com/13573011/29321279-9627110c-81e2-11e7-81ab-b9991b028f85.png)

What's the problem???

NycuRO:
![image](https://user-images.githubusercontent.com/13573011/29321330-c3b67bb2-81e2-11e7-9833-feb867170e6e.png)

And on Single

??

asyven:
try to place blocks above tnt

sero583:
I did that on a flat world and place TNT over one bedrock at y=1 @NycuRO @asyven

SupremeMortal:
Has anyone looked into the code. It can't be that hard to fix.

sero583:
[...]

PR #1864, opened 2017-08-15
Fix for #1861

sero583:
Small explanation, I reduce the call to Block.getId(); in the way to store it as an int don't really know if this affects much probably not but I like it in the way and why do I don't check the explosion resistance of the block in this fix? Because it's obvious for me these two blocks are not destroyable thats why I block in that way further calculations, which is imo performance improvement. Sorry if formatting isn't in the way it should be. (@SupremeMortal , @NycuRO)

Edit: Oops title does not link to issue so here #1861.

### DIFF
--- a/src/main/java/cn/nukkit/level/Explosion.java
+++ b/src/main/java/cn/nukkit/level/Explosion.java
@@ -90,8 +90,8 @@ public class Explosion {
                                 break;
                             }
                             Block block = this.level.getBlock(vBlock);
-
-                            if (block.getId() != 0) {
+                            int id = block.getId();
+                            if (id != 0 && id != Block.OBSIDIAN && id != Block.BEDROCK) {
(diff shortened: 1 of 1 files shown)


sero583:
Update/Statement: Well I don't think this quickfix is reliable for the future, since probably new blocks will be added and storing them in that way is maybe not the best way because of updating this could be forgotten probably.
Searching for a better method to fix this now.

sero583:
Closing this due to not reliable way and also very foolish way.

PetteriM1:
Please reopen this

sero583:
@PetteriM1 Why?

PetteriM1:
Because someone has to fix this

sero583:
@PetteriM1 Ofcourse it has to be but thats not the suitable fix I guess. Also the issue is open for it and Nukkit is currently unfortunately inactive.

PR #1870, opened 2017-08-20
Removed redundant message

XxDXM-hippiexX:
Removed the "Player data not found for "Steve", creating new profile" when save-player-data is set to false.


PeratX:
I don't think your modification is correct

sero583:
@PeratX could you merge some finished important PR's for example setMotion(); fix. It seems to be @Snake1999 is pretty inactive and can't merge it.

PR #1877, opened 2017-08-24
Fix block breaking

BukkitPlaysMC:
This is needed, otherwise Nukkit will think the player is using fast break multiple times and it will not break the block.


NycuRO:
:D

PikyCZ:
I mean break time is 0.15?

PikyCZ:
@PeratX  yu can close this PR

PR #1881, opened 2017-08-26
Fix BlockPlaceEvent not getting called

BukkitPlaysMC:


PeratX:
There is nothing wrong with original code.

PR #1884, opened 2017-08-27
final fix for crafting

Creeperface01:
hopefully fixed all recipes


sero583:
Hopefully, does this means it's not tested?

Creeperface01:
I tested it, but not all of 400 recipes

NycuRO:
@Creeperface01 In my branch of 1.2 with packets, it's all recipes in .json 👍

PR #1890, opened 2017-09-03
Update EntityEventPacket.java

azon0320:
Add entity eat food animate


xISRAPILx:
This will add a constant ... We still need to send the package.

azon0320:
Server will receive when player is eating food ,because player eat mutely in server now

PeratX:
Any tests to support this?

azon0320:
Yes, you can register a listener to receive this packet

sero583:
The PR creator should provide more info's etc. to prove it, for example a video. And yes with that you are meant, @azon0320.

PR #1900, opened 2017-09-21
Update to 1.2.0

maxxie114:


Goodstuff4UonYT:
@maxxie114  1.2 is a big update. Only changing the protocol info does not work.

maxxie114:
It will make the server joinable from 1.2.0, yes, 1.2 is a big update, and it needs a lot more development.  Nukkit is too outdated, they don't have a lot of the New Items.

Goodstuff4UonYT:
Did you test this? :'D I am pretty sure nothing will work.

SupremeMortal:
It will not make the server join able. The way packets are sent has had a protocol breaking change so it's best you wait.

PeratX:
Wait for updates by @yescallop

PikyCZ:
I mean by @Creeperface01  :P

Issue #1869, opened 2017-08-17
Disable logs


PetteriM1:
Please add option to disable logs or tell how can I do it in my own nukkit fork.

PetteriM1:
Does anyone know?

Creeperface01:
override MainLoger class :D

PR #1908, opened 2017-10-01
Allows for the log level to be changed

spammy23:
Using the command line parameter `--verbosity` or `-v`
The log level is case-insensitive.

For example, 
`java -jar Nukkit.jar -v none`
`java -jar Nukkit.jar --verbosity emergency`


spammy23:
Resolves #1869

boybook:
Is it something wrong? If we turn on debug, we will just see debug messages.

boybook:
It should be >= instead of <=.

PR #1926, opened 2017-11-03
Working Multiversion, @CreeperFace01 's crafting fix, 1.2.5 compatibility

kvetinac97:
Now, you can connect with Minecraft PE 1.1.0-5 to Nukkit!
Everything except for containers transactions (taking items from chest etc. - bugging) and equipping items to other hotbar slots that 1 -> 1...


SupremeMortal:
Can multiversion not be maintained in a separate project?

src/main/java/cn/nukkit/utils/Utils.java
```diff
@@ -209,6 +209,9 @@ public static void reverseArray(Object[] data) {
     }
 
     public static int toInt(Object number) {
+        if (number instanceof Double){
+            return Double.valueOf(Math.round((double) number)).intValue();
+        }
         return (int) number;
```
SupremeMortal: What about floats? The original method was more effective because casting a float to a double and then rounding cause no precision to be lost.

kvetinac97: If you put a Double or Float to (int) something, it will throw error.

You can convert Integer, double, float... to int. But not Double. That’s why I changed it

SupremeMortal: This is probably the best way. It will throw an Arithmetic exception if a double or long is larger than `Integer.MAX_VALUE`
```
    public static int toInt(Number number) {
        return Math.toIntExact(Math.round(number.doubleValue()));
    }
```

liach:
Is nukkit dead?

SupremeMortal:
@liach https://github.com/NukkitX/Nukkit

Issue #1948, opened 2017-11-22
Power enchantment not increasing damage correctly


ScraM-Team:
### Issue Description
Arrows fired from bows with the Power enchantment don't seem to do as much damage as they are supposed to.  Zombies don't die instantly like they do in PC Minecraft when struck with a Power 1000 bow.

### Steps to Reproduce the Issue
Give a player a Power 1000 bow
 bow.addEnchantment((new EnchantmentBowPower()).setLevel(1000));
Shoot a Zombie.
It doesn't die.

I think there is at least one problem with this logic:
ItemBow.java:62  damage += 0.25 * (bowDamage.getLevel() + 1);

I think it should be
damage += damage * 0.25 * (bowDamage.getLevel() + 1);

per https://minecraft.gamepedia.com/Bow#Enchantments

### OS and Versions
* Nukkit Version: https://github.com/Nukkit/Nukkit/commit/b04061b912ec16c4c9c753f2e475f6e2e9514402
* Java Version: 1.8.0_121

* Client Configuration: 

| Item | Value |
|:----:|:-----:|
| Client Edition | Android | 
| Client Version | 1.2.3 |

PR #1956, opened 2017-11-30
Fixes for arrow damage from enchanted bows

ScraM-Team:
This fixes two minor issues with calculating damage for arrows fired from enchanted bows.


SupremeMortal:
Weird because the original calculation is the one pulled from vanilla. Maybe there is a mis-calculation elsewhere in the damage calculation code.

ScraM-Team:
I used the formula from the wiki (https://minecraft.gamepedia.com/Bow#Enchantments).  Oh, I forgot to mention this fixes issue #1948 .

PR #1963, opened 2017-12-07
The great multiversion update (1.1.0 - 1.2.6)

kvetinac97:
This update brings multiversion to Nukkit.
You can join with any version between 1.1.0 and 1.2.6.
100% working and tested are: 1.1.3, 1.2.3, 1.2.5 and 1.2.6.

PR also contains:
 - Fix of air equipping leggings
 - Fix not being able to wear skulls (even when they are helmets)
 - Fixing CommandEnums for 1.2
 - Crafting fix by @Creeperface01


WetABQ:
wowwwwwwwwwwwwwwwwwwwwwwwwwwwww 
good job!!!!!!

zzz1999:
This pull requests are supposed to merge as soon as possible.

SupremeMortal:
There is no one to merge them since all of the people with repository access have become inactive.

diamond-gold:
In this PR there's a bug, if you quit far away from spawn, you will be stuck in building terrain until you are teleported near spawn

WetABQ:
The client crashes when the player opens the backpack.
I can see the backpack has a lot of grids
But with just one click it will recove
There are usually a few stones in the backpack
(1.1.3 version)1.2 version does not appear this situation
Thanks

NycuRO:
@kvetinac97 Fix errors and that will be future 👍

PeratX:
This path is not right in my opinion. You should design a more extensible multiversion protocol framework.
This implementation is not extensible for newer protocols.

SupremeMortal:
Better option would be to have separate classes and instantiate depending on the version received.

kvetinac97:
Yes, when 1.3 will come out, there will be need for a different initiating and handling

Creeperface01:
maybe my implementation is better

kvetinac97:
Any method is better than this. I’ll work on it later. And @Creeperface01, I don’t think implementing multiversion via plugin is better than have it in Nukkit.

PeratX:
Non-extensible multiversion framework will never be a part of Nukkit. I have informed you.
This implementation can be a part time toy for some one who needs it.

kvetinac97:
When 1.3 will come, I’ll rework it to be extensible

EnderWarrior:
How to download?

kvetinac97:
http://circleci.com/gh/kvetinac97/Nukkit/

NycuRO:
Can you import this branches/PR's to this PR?

https://github.com/Nukkit/Nukkit/pull/1956

https://github.com/Creeperface01/Nukkit/commits/features%26fixes

RoofyManStar:
@kvetinac97 Why you not create one plugin for that addon?

Master branch using only protocol.
feature branch using add and features.

WetABQ:
1.1 version of the game player to eat food can not be restored to fullness, and can not use the workbench synthesis.

WetABQ:
The command needs to be entered multiple times to respond

kvetinac97:
at least something fixed 😄

src/main/java/cn/nukkit/Server.java
```diff
@@ -923,7 +923,8 @@ public void sendFullPlayerListData(Player player) {
     }
 
     public void sendRecipeList(Player player) {
-        player.dataPacket(CraftingManager.packet);
+        player.dataPacket(player.getProtocol().equals(PlayerProtocol.PLAYER_PROTOCOL_113) ?
+                CraftingManager.packet113 : CraftingManager.packet);
```
Creeperface01: wtf proč používáš equals když protokol je číslo? 😆

kvetinac97: Já to mám jako enum

Creeperface01: no však :D odkdy se enum porovnává pomocí equals

SupremeMortal: Enums are final Objects so you can compare them with == instead of running equals().

src/main/java/cn/nukkit/inventory/ShapelessRecipe.java
```diff
@@ -19,10 +19,22 @@
...
     private final List<Item> ingredients = new ArrayList<>();
 
+    private int recipeProtocol = 130;
+
     public ShapelessRecipe(Item result) {
         this.output = result.clone();
     }
 
+    @Override
+    public boolean isCompatibleWith(int protocolVersion) {
```
NycuRO: Srsly :/

Qitu:
Sometimes, I try to.change player's gamemode, but it always give me a strange result, for example, viewer gamemode change to Survival gamemode, but he still can fly, Survival to Create, cannot fly......

src/main/java/cn/nukkit/inventory/ShapelessRecipe.java
kvetinac97: Yes, because 1.1 will crash if you send it recipes from 1.3

kvetinac97: 1.2*

src/main/java/cn/nukkit/Server.java
kvetinac97: Lel :smile: I’ve been using them for 2 years and I still didn’t know that

src/main/java/cn/nukkit/inventory/ShapelessRecipe.java
NycuRO: Yes bro, but use one boolean function with true and false. Why true any time?

Qitu:
Float text entity cant be seen in 1.2

NycuRO:
@kvetinac97 Can you create this PR to be a plugin? Thanks..

NycuRO:
Can you implement xbox auth?

Snake1999:
However it should be considered to design a more extensible application interface, which allows for a further update.

NycuRO:
Versions < 1.1.3 freeze server.

NycuRO:
if player < 1.2 get fireworks, inventory get buggy.

kvetinac97:
I will be making an extensible version, starting from today. Enums will stay, but there will be some MainEnums (1.1, 1.2) and ExtensingEnums (1.2.5), so packers will be decoding it as protocol.isMainEnum(PROTOCOL_130), which will be true for 1.2.5...

Also, packet decoding that is different in other versions will be using switch rather than if else statements. The main goal will be to make the update able to merge

PR #1974, opened 2017-12-22
bug fix: diamond pixel can't mine obsidian

reezhu:
time of golden tool and diamond tool mismatch


SupremeMortal:
The way you have _fixed_ it is wrong. Gold has a larger multiplier than diamond. This needs to be changed on the specific block.

reezhu:
find that, but the bug still remains

PR #1979, opened 2018-01-26
The great & extensible Multiversion update!

kvetinac97:
This update is compatible with versions: 1.1.0 - 1.2.9
----------------------------------------------------------
What's new:
 - All packets now use PlayerProtocol enum for decoding/encoding/pid
 - Each PlayerProtocol has its own number (x.x.X), main number (x.X.x) and protocolInfo class
 - When adding a new Protocol, you just have to add new ProtocolInfoXXX class and a new PlayerProtocol enum. Then, just use it where you want.
 - All packets that are decoded/encoded automatically update for all available protocols, so when you add a new protocol, you don't have to worry if old packets will still work
 - Player class recognizes packets by name (to prevent id conflicts from different versions)!

The update contains also:
 - Fix of air equipping leggings, allowing skull to be helmet
 - Fix of command enums
 - Crafting fix by @Creeperface01 
 - Fireworks! (thanks @Creeperface01)


SupremeMortal:
There's still if statements all over the packet encode/decode methods. I would hardly call that extensible.

kvetinac97:
When you look at Steadfast2, there are also lots of if statements. And when you will add a new protocol, everything will work with these if statements. New protocol won’t be 113 and will be greater than 130

sero583:
Multiversion should be seperated from Nukkit in my opinion.

WetABQ:
I think you should implement an internal abstracted packet processing system instead of being scattered around Nukkit, with no changes internally, and only need to be visualized as concrete client-side packets when sent to the player. In reading the data packet should be the first package to convert abstract data to deal with the data, so that most of the internal structure of Nukkit will not be changed.
This is their ideal structure.
However, the project manager for other reasons can not write such a system.

SupremeMortal:
You're going to have to revert that commit. Those changes never made it to release. It was just a protocol bump 🤣

lsw46:
Maybe it would be better to make this PR a plugin

SupremeMortal:
All that was added was the ServerChunkTickRange in StartGamePacket.

kvetinac97:
As of NukkitX updates were merged into master, there are conflicts that I, as for now, will not be resolving, so unless it will be needed, by from me 😀

PR #2012, opened 2018-05-19
new repository

Creeperface01:


Creeperface01:
yes I saw that, but it's almost empty at the moment. wouldn't be better to work on one project than writing whole software from scratch? NukkitX rewrite is also completely new software so I don't think it's necessary to create another software and split the community. And as I said NukkitX rewrite is in the advanced development phase and I think it's good enough so it's not worth it to create new software. Or has your software any other pruporses that couldn't be done in nukkitX?

SupremeMortal:
@Snake1999 You can check it out here: https://github.com/NukkitX/Nukkit/tree/rewrite
We already have most of the core code written and support for the new architecture of Minecraft with Block States, Data Layers and runtime IDs.

SupremeMortal:
This repository confuses a lot of new comers to Nukkit so wouldn't it make sense to merge? The current README is also outdated and the download link does not work.

PR #2014, opened 2018-05-23
fix short read bug

reezhu:
this bug may cause server not able to read long nbt tag,
in java << has higher propriety than &


sero583:
@reezhu This was already done by @w-gao...

